PR #37, opened 2023-10-17
Fixed small typo in .env file to prevent needless debugging for newcomers.

BHenkemans:
Just merge this, it is a really small fix but main is protected so I could not commit directly :/


CodeNamedRobin (approved):
Much PR, very code, such fix. Looks good to me!

PR #55, opened 2023-10-17
Fix menubar not updating when changing locale

### LINKED ISSUE #54: Changing the locale does not change the menubar
Changing the locale does change the whole page, but not the menubar, because of this: https://github.com/intlify/vue-i18n-next/issues/1396

SuperVK:
fixes #54


noelvandevrie (approved):
mooi

PR #59, opened 2023-10-17
Added translations, as well as other translation improvements

### LINKED ISSUE #50: Missing Translations
There seem to be some missing translations on the homepage.

SuperVK:
- Added translations where they were still missing, only place where it still is missing is in the Recent transactions, but those are hardcoded either way so I'm expected them to be overhauled.
- Renamed a bunch of in-code cases of "Saldo" cause that is not an English word :)
- renamed some things that should be not c_ in the balance view, but should be improved up on more in a translation file overhaul.

Fixes #50

### DIFF
--- a/src/components/BalanceComponent.vue
+++ b/src/components/BalanceComponent.vue
@@ -1,7 +1,7 @@
 <template>
   <CardComponent
-      :header="$t('c_currentSaldo.saldo')"
-      :action="showOption ? 'Increase Online' : ''"
+      :header="$t('c_currentBalance.balance')"
+      :action="showOption ? t('c_currentBalance.Increase balance') : ''"
       routerLink="balance"
   >
     <div class="body">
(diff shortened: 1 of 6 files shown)


src/components/BalanceComponent.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <CardComponent
-      :header="$t('c_currentSaldo.saldo')"
-      :action="showOption ? 'Increase Online' : ''"
+      :header="$t('c_currentBalance.balance')"
```
CodeNamedRobin, line 3: If you're importing t from vue-i18n, you can leave out the $ right? It doesn't need to use the global definition then.

CodeNamedRobin (changes requested):
One minor thing.

CodeNamedRobin (approved):
LGTM!

PR #64, opened 2023-10-20
Added favicon and title to <head>

### LINKED ISSUE #62: Implement proper html title and favicon / metadata
There needs to be a <title> tag and a favicon such that it doesn't keep saying Vite app with a vue favicon.

SuperVK:
Fixes #62


CodeNamedRobin (approved):
I'm approving this, but we should put the "get" functions for different roles somewhere else. Right now it feels a bit chaotic. I will open an issue for this.

PR #74, opened 2023-10-25
Feature/error handling toasts

### LINKED ISSUE #18: Incorporate error handling and error toasts
PrimeVue already has a very nice implementation of toasts, https://primevue.org/toast/, so we want to use these to implement the error handling. This would also include proper error handling using .then and .catch for every async function.

CodeNamedRobin:
Fixes #18.
Now using proper toasts from PrimeVue.
Also using proper .then and .catch blocks instead of if/else blocks.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,9 +1,11 @@
 <script setup lang="ts">
 import { RouterView } from 'vue-router';
+import Toast from "primevue/toast";
 
 </script>
 
 <template>
+  <Toast />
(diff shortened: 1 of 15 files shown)


SuperVK:
Isn't the error for incorrect credentials also included? Right now that silently dropped. Or is that out of scope of this PR?

SuperVK (approved):
Nvm, CORS is annoying

PR #94, opened 2023-12-12
Fixing several smaller problems in the add product modal

### LINKED ISSUE #89: Decimals points not possible when adding product
When adding/creating a new product, only integer values are possible because you cannot type a "," or ".".

### LINKED ISSUE #90: After creating new product doesn't close down modal
When clicking save on a product, a confirmation appears, but it doesn't close the button. Furthermore the confirmation toast has an untranslated file
image

And doesn't update products (only after reload)

BHenkemans:
Fix GH-89, GH-90.


CodeNamedRobin (approved):
LGTM

PR #125, opened 2024-02-22
Upgraded to primevue 3.49.0 and fixed styling issues

SuperVK:
Primevue moved their styling from primevue to their theme files (https://github.com/primefaces/primevue/issues/5249#issuecomment-1937899018). This caused us not to be able to upgrade to primevue 3.48.1. I have updated our theme files with the files from the primevue-sass-theme repository. 

If somebody could please check there are no visual changes that would be appreciated!

Also I upgraded our vite because that contained a vulnerability.


CodeNamedRobin (approved):
LGTM

PR #132, opened 2024-03-17
Redesign of the login view

CodeNamedRobin:
Redesigns the login view to among other things bring back the floating beers. Also splits the UI in local login and login with GEWIS.


SuperVK (approved):
LGTM!

PR #155, opened 2024-04-15
Fix issues concerning profile page

### LINKED ISSUE #151: [BUG] Profile page is improperly implemented
One issue for now, but maybe want to split this up:
- [x] You cannot toggle intensive data processing on the profile page
- [x] You have the option to change your name, but this is throws an 403
- [x] You can generate API keys, which only admins should be able to do.
- [x] Pin codes should be hidden by a password field instead of plaintext
- [x] External accounts cannot see how they are supposed to log into the POS, the sudosos number is not shown

CodeNamedRobin:
Fixes #151.


SuperVK:
Are you placing the UserInfoCompomenet in the user overview in this PR? Or should I move that in the UX PR?

CodeNamedRobin:
@SuperVK the UserInfoComponent is placed in the profile page in this PR.

src/views/ProfileView.vue
```diff
@@ -1,18 +1,24 @@
...
       </div>
-      <div class="col-12 md:col-6">
+      <div class="col-6 md:col-6" v-if="isAdmin">
         <ChangeApiKeyComponent />
-      </div>
-      <div class="col-12 md:col-6">
-        <ChangeInfoComponent />
+        <div class="flex flex-row mb-2 align-items-center">
+          <h3 class="mr-3">{{ $t('profile.individualDataAnalysis') }}</h3>
+          <InputSwitch v-model="dataAnalysis" @update:modelValue="handleChange" />
```
SuperVK, line 20: Why is this only allowed when you are an admin?

SuperVK:
General thing: Some cards are disabled when you are not allowed to change them, whilst others are just plain hidden. Maybe we should make a clear distinction in this and choose one or the other?

src/views/ProfileView.vue
CodeNamedRobin: fixed!

SuperVK:
image
Found one more, but not in your code so can't comment on it, but updating pin says updating password

SuperVK:
Furthermore, the isLocal boolean seems to break on refresh: (this is a local account that shows correctly after login, but refreshing disables the password field again)
image

SuperVK:
Also: For readability I would toggle each component in ProfileView.vue, not inside the individual components like is done now.

CodeNamedRobin:
@SuperVK I think it should be mainly good now. About disabling and grayscaling I think we might want to have a separate discussion, as it right now is in line with what is currently on sudosos.gewis.nl. Maybe after release have a look?

src/locales/en.json
```diff
@@ -285,12 +287,5 @@
     "passwordUpdated": "Password successfully updated",
     "updatedUserInfo": "Succesfully updated user info.",
     "userInfoUpdated": "User information successfully updated"
-  },
-  "contact": {
```
SuperVK: I am not sure how it got removed, but the contact translation were quite useful...

SuperVK (changes requested):

src/locales/en.json
CodeNamedRobin: How did this happen lol, will fix

CodeNamedRobin: Fixed!

SuperVK (approved):

PR #156, opened 2024-04-16
Fix several minor fine overview issues

### LINKED ISSUE #118: Fine issues
- [x] While applying the dates on fine overview, some users' names don't appear. I suspect this is because of GewisUserResponse and UserResponse conflicts.
- [x] Selection should disappear after notify or hand out.
- [x] There is no feedback on notify.
- [x] When opening the datepicker for the reference dates, the time defaults to the current minute, but resets to 12 AM

CodeNamedRobin:
Fixes #118.

### DIFF
--- a/src/components/Mutations/MutationsTableComponent.vue
+++ b/src/components/Mutations/MutationsTableComponent.vue
@@ -91,7 +91,7 @@ import type {
   TransferResponse
 } from "@sudosos/sudosos-client";
 import { onBeforeMount, type Ref } from "vue";
-import { onMounted, ref } from "vue";
+import { ref } from "vue";
 import MutationModal from "@/components/Mutations/MutationModal.vue";
 import {
   type FinancialMutation,
@@ -100,7 +100,6 @@ import {
   parseTransaction,
   parseTransfer
 } from "@/utils/mutationUtils";
-import { formatPrice } from "@/utils/formatterUtils";
 import "primeicons/primeicons.css";
 import { useI18n } from "vue-i18n";
 
@@ -138,7 +137,6 @@ const selectedMutationType = ref<FinancialMutationType>(FinancialMutationType.TR
 const mutationShow = ref<boolean>(false);
 const totalRecords = ref<number>(0);
 const isLoading: Ref<boolean> = ref(true);
-const skeleton = ref(new Array(10));
 const { t } = useI18n();
 const rows: Ref<number> = ref(10);
 onBeforeMount( async () => {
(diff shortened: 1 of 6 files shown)


src/components/Mutations/MutationsTableComponent.vue
```diff
@@ -100,7 +100,6 @@ import {
   parseTransaction,
   parseTransfer
 } from "@/utils/mutationUtils";
-import { formatPrice } from "@/utils/formatterUtils";
```
SuperVK, line 103: Even though the IDE says this is unused, it is in fact in the <template> and removing this break everything:
image

src/views/FineView.vue
```diff
@@ -268,14 +290,38 @@ const handoutFines = async () => {
...
+        detail: t('successMessages.finesHandedOut'),
+        life: 3000,
+        severity: 'success',
+      });
+    })
+    .catch((err: AxiosError) => handleError(err, toast))
+    .finally(() => selection.value = []);
+};
+
+// NOTE: This is a temporary fix, until PrimeVue 3.52, where this issue should be fixed in the library
```
SuperVK: I propose we do not do this fix, just upgrade when 3.52 releases. It is not that much of an issue and I expect 3.52 to be released before 22 may.

src/views/FineView.vue
```diff
@@ -254,10 +264,22 @@ const openHandoutEvent = async (eventId: number) => {
 };
 
 const notifyUsers = async () => {
+  isNotifying.value = true;
   await apiService.debtor.notifyAboutFutureFines({
     userIds: selection.value.map((item: any) => item.id),
     referenceDate: secondDate.value?.toISOString() || new Date().toISOString()
-  });
+  })
```
SuperVK, line 269: Why are you both awaiting, and using .then()? https://dev.to/masteringjs/using-then-vs-async-await-in-javascript-2pma

src/views/FineView.vue
```diff
@@ -191,20 +196,25 @@ const handlePickedDates = handleSubmit(async (values) => {
...
       ...item,
+      // @ts-ignore
+      gewisId: (userStore.users.find((u: UserResponse) => u.id === item.id) as UserResponse).gewisId || undefined,
       fullName,
       // Assign first and last balance based on the first and second items in the balances array
       firstBalance,
       lastBalance: secondBalance,
     };
-  });
+  }).filter((u: any) => !deletedUsers.includes(u.id) && activeUsers.includes(u.id));
```
SuperVK, line 215: Why is this not fixed in the backend but in the frontend?

SuperVK (changes requested):

src/views/FineView.vue
CodeNamedRobin: I want to do both, and for now the bug is not fixed in the backend yet.

SuperVK (approved):
LGTM!

PR #160, opened 2024-05-01
fix: correct redirection after logging in with GEWIS

### LINKED ISSUE #141: [BUG] Login with GEWIS does not redirect to the HomeView
Basically the title. Login is succesful as per the network response, but the user is not redirected to the Home Page.

SuperVK:
fixes #141


CodeNamedRobin (approved):
LGTM!

PR #177, opened 2024-05-21
fix: SingleUserView broke on not assigning userType in code

### LINKED ISSUE #176: [BUG] SingleUserView is broken
I don't know what it does, but "TypeError: i is null" on load and the form does not like any updates.

CodeNamedRobin:
Fixes #176


SuperVK (approved):
LGTM!

PR #178, opened 2024-05-21
fix: several issues on product containers view

### LINKED ISSUE #171: Pagination in Product Overview not doing anything
The pagination buttons and dropdown seem to be disabled, because clicking does not do anything?

image

Furthermore, it is also good practice to include some information about the records you're showing. Something like `Showing 1-5 of 312`, so the user has some idea where he is in the list.

### LINKED ISSUE #172: Product images in containers not scaling properly
WIDE WINE

image

CodeNamedRobin:
Fixes #171 
Fixes #172


SuperVK:
image
Wine is still wide when opening the product

SuperVK (approved):

PR #180, opened 2024-05-21
fix/enhancement: BalanceComponent not crashing and loading skeletons

### LINKED ISSUE #166: Balance component shows 0 as balance when still loading
Might be better to either show a placeholder or a loading animation to indicate to the user that their balance is still being fetched. If somehow something is not working, it will show 0 euros.

image

### LINKED ISSUE #170: Topping up with a negative amount causes the component to crash
If you type "-10" in the top up field, the component crashes.

image

CodeNamedRobin:
Fixes #170 
Fixes #166 
Also, for future reference, Dinero does not like it when any non integer is given (1005.0000001 is not allowed)

### DIFF
--- a/src/components/BalanceWithTopupComponent.vue
+++ b/src/components/BalanceWithTopupComponent.vue
@@ -27,7 +27,7 @@
             }" @input="(test) => {
                 setFieldValue('Top up amount', test.value as number, errors['Top up amount'] != undefined);
                 setTouched(true)
-              }" :placeholder="$t('balance.Price')" inputId="amount" mode="currency" currency="EUR" locale="nl-NL" />
+              }" :placeholder="$t('balance.Price')" inputId="amount" mode="currency" currency="EUR" locale="nl-NL" :min="0"/>
           </div>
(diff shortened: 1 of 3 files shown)


src/components/BalanceWithTopupComponent.vue
```diff
@@ -20,14 +32,27 @@
...
+              currency="EUR"
+              locale="nl-NL"
+              :placeholder="$t('balance.Price')"
+              :min="0.0"
+              :min-fraction-digits="0"
+              :max-fraction-digits="2"
+              v-model="topupAmount"
+              inputId="amount"
+              @input="
+                (test) => {
```
SuperVK: maybe nice to actually change the "test" variable I once introduced when testing but never changed to "value" or smth

SuperVK (approved):
LGTM!

src/components/BalanceWithTopupComponent.vue
CodeNamedRobin: Fixed, named to data

PR #179, opened 2024-05-21
fix: Several localisation issues.

CodeNamedRobin:
Fixes #175 
Fixes #173
Fixes #165 
Fixes #158

PR #181, opened 2024-05-21
enhancement: distinguish between transactions

### LINKED ISSUE #169: No distinguishment between own tranactions and transactions created for others
The list of transactions includes many records, but for BAC members, it is impossible to see which transactions are for myself and which transactions I created during a borrel and are thus for someone else. The only way to see this now is to click every individual transaction. Maybe make the transactions created for someone else gray, as they are less important to me (but still important to show).

image

CodeNamedRobin:
Fixes #169 .

### DIFF
--- a/src/components/Mutations/MutationsTableComponent.vue
+++ b/src/components/Mutations/MutationsTableComponent.vue
@@ -8,7 +8,7 @@
       lazy
       @page="onPage($event)"
       :totalRecords="totalRecords"
-      :rowClass="getOtherTransaction"
+      :rowClass="isSomeoneElsesTransaction"
     >
       <Column field="moment" style="width: 30%" :header="$t('transactions.when')">
         <template #body v-if="isLoading">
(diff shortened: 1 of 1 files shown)


src/components/Mutations/MutationsTableComponent.vue
```diff
@@ -8,25 +8,27 @@
...
           <span class="hidden sm:block">{{ mutation.data.moment.toDateString() }}</span>
-          <span class="sm:hidden">{{
-            mutation.data.moment.toLocaleDateString('nl-NL', {
-              dateStyle: 'short'
-            })
+          <span class="sm:hidden"
+            >{{
+              mutation.data.moment.toLocaleDateString('nl-NL', {
+                dateStyle: 'short'
+              })
```
tomudding, line 23: Just a FYI, this will likely cause conflicts with GH-179 when merging due to the formatting changes.

src/components/Mutations/MutationsTableComponent.vue
```diff
@@ -215,6 +230,12 @@ function getDescription(mutation: FinancialMutation) {
     }
   }
 }
+
+const getOtherTransaction = (data: FinancialMutation) => {
```
SuperVK: This could use a better name I think

SuperVK (approved):

src/components/Mutations/MutationsTableComponent.vue
CodeNamedRobin: Fixed!

src/components/Mutations/MutationsTableComponent.vue
CodeNamedRobin: True, although I'll fix that when rebasing.

PR #185, opened 2024-05-22
Templates for issues and pull requests

SuperVK:
This will add templates for issues and pull requests to make the workflow a lot cleaner.


JustSamuel (approved):
Just an FYI, I also commonly see a 

📝 Documentation
🛠 Tooling

issue template, but I don't think we will need them.

Issue #188, opened 2024-05-28
[Refactor] Autoimport components


SuperVK:
Problem: The components/views are a mess right now, the names have no common convention. They need to be grouped properly, such that navigation comes easy again.

Steps:
- [ ] Use https://github.com/unplugin/unplugin-vue-components to auto-import components, such that we do not have to import the components everytime, including primevue.
- [x] Choose whether we want to create folders per components, or do everything top level and follow [the official vue recommendation](https://vuejs.org/style-guide/rules-strongly-recommended.html#tightly-coupled-component-names)

SuperVK:
unplugin is not too relevant right now, if we need it later we can re-open the issue

PR #191, opened 2024-05-28
style: autoimport components

SuperVK:
Switch to auto importing components and remove all existing imports

# Description
Auto importing components using https://github.com/unplugin/unplugin-vue-components

Fixes the first step of #188 

## Types of changes
- Code style improvement


SuperVK:
I will look at this again when moving into the monorepo

PR #193, opened 2024-05-30
Hotfix/update common

SuperVK:
See https://github.com/GEWIS/sudosos-frontend-common/pull/8 
Updates to include this update


JustSamuel (approved):
pls in the future add the commit hash to github hosted packages because an update to the lock without changes to the package.json is bad.

PR #204, opened 2024-06-08
Improved POS login

SuperVK:
Improved the styling of the keypad, and improved and changed rendering of the banners

# Description
The styling of the keypad has now been changed to remove the continue button in favor of being able to tap the userid and passcode field. As well as updated buttons.

Furthermore the banners rendering have changed. First of all, banners that would not fit the size are now in their original aspect ratio, and are rescaled to fit in the slot, with the excess being filled by the poster background from TVPC/aurora. Furthermore the active checkbox is now to override when set to false, instead of override when set to true. And banners without an image are not rendered anymore, but instead skipped.

Finally there is now the option to see a little bit of information on how to login into the POS, for new GEWIS'ers or GEWIS who are unfamiliar to SudoSOS.

image
image

Finally the POS and dashboard now both respect banner duration to be in seconds, not in milliseconds :)

No related issues.

## Types of changes

- Design changes


JustSamuel:
Single Responsibility: I have almost no remarks on the banner side, but some small on the extra login thingy, so maybe these should have been 2 PR's.

apps/dashboard/src/components/BannerItemComponent.vue
```diff
@@ -64,8 +64,8 @@ defineProps<{
 const { t } = useI18n();
 
 const displaySeconds = computed(() => {
-  return `${(banner.value.duration / 1000).toLocaleString()}
-   ${banner.value.duration / 1000 != 1 ? t('banner.seconds') : t('banner.second')}`;
+  return `${(banner.value.duration).toLocaleString()}
+   ${banner.value.duration != 1 ? t('banner.seconds') : t('banner.second')}`;
```
JustSamuel: I know this isn't your original code; but we should use pluralization instead https://vue-i18n.intlify.dev/guide/essentials/pluralization

JustSamuel (changes requested):
In general I agree, but I do think that the "how to login" is cramped between the keypad and the banner. Maybe move it some where else? bottom left in the form of an icon + image?

SuperVK:
image
I was thinking of 3 or 4 myself first. Do you mean 1 or 2 in for the bottom? And then like a question mark or exclamation mark icon?

JustSamuel:
> I was thinking of 3 or 4 myself first. Do you mean 1 or 2 in for the bottom? And then like a question mark or exclamation mark icon?

I think 1 would be best; but maybe show the options to some different people?

SuperVK:
image
image
image
image

JustSamuel (approved):
LGTM

PR #206, opened 2024-06-12
Development quick start

SuperVK:
Added development quick start instructions

# Description
Added instruction for Codespaces set up and local setup, as well as improvements to the devcontainer that is used when using codespaces. 

-

## Types of changes

- Documentation improvement


README.md
JustSamuel, line 1: Contributing is usually always last.

And correct me if I'm wrong, but with NPM codespaces / subprojects you dont need to CD right? iirc you can simply npmi from root and run commands from there.

JustSamuel (approved):
Small remark

README.md
SuperVK: Yeah true, but I'd rather have people see first how to open an issue, and then how to solve them yourself.

And `npm i` works from root, but `npm run dev` only works from the specific directories in my case (or `npm run dev --workspace=sudosos-pos-vue-3` (for example) , also works, might be worth it)

PR #213, opened 2024-06-15
fix: Prevent deleted users from breaking user mapping on FineView

### LINKED ISSUE #212: [BUG] Clicking Apply Dates on Fine Overview Gives an Error

Chrome, Firefox, Microsoft Edge

Dashboard

### Current behaviour

When clicking apply on the fine overview page, nothing happens and an error is given in the console about gewisId being undefined.

### Desired behaviour

When clicking apply, a list of users should be given who are eligible for a fine based on the reference dates.

### Steps to reproduce

1. Login on the dashboard using an account with permission to hand out fines.
2. Go to the fine overview
3. Select two dates
4. Click apply
5. Open console

### Other information

_No response_

CodeNamedRobin:
# Description
Fixes issue where deleted and inactive users were considered in a mapping for the FineView and breaking the mapping when trying to find this user.

Fixes #212.

## Types of changes


apps/dashboard/src/main.ts
SuperVK, line 1: Why not just roll back this file, nothing changed, also not relevant to the bug your fixing right?

apps/dashboard/vite.config.ts
```diff
@@ -16,6 +16,11 @@ export default defineConfig({
   optimizeDeps: {
     exclude: ['ToastService'],
   },
+  // build: {
+  //   rollupOptions: {
+  //     external: ['@sudosos/sudosos-frontend-common'],
+  //   },
+  // },
```
SuperVK: Why?

package-lock.json
SuperVK, line 1: Why? Either put this in the .gitignore with a good reason, or not remove it (also seperate PR cause it does not seem related to the bug you are fixing)

apps/dashboard/vite.config.ts
CodeNamedRobin: Fixed

package-lock.json
CodeNamedRobin: Fixed.

apps/dashboard/src/main.ts
CodeNamedRobin: True, rolled back.

SuperVK (approved):

PR #220, opened 2024-06-19
ci: monorepo docker

### LINKED ISSUE #214: Issue with github actions and package-lock
Right now, when using this repo in development, `npm install` will take the package-lock at the root as truth. Installing packages from local files, like is supposed to happen, **furthermore, npm install will no longer use or overwrite the package-locks in the applications**. However (comma) when running `docker build` with the context of the respective apps, the docker will use the package-lock from the apps, which still contains the old GitHub repositories as installation source for the common library (because they were never overwritten when running `npm install` in development). Thus, right now, updating the common library does not actually do anything.

To fix this, we should probably look at something like this, where the context is still the whole repository: 
https://stackoverflow.com/questions/50215553/how-to-build-a-docker-image-from-a-nodejs-project-in-a-monorepo-with-yarn-worksp
https://github.com/docker/build-push-action?tab=readme-ov-file#inputs

SuperVK:
Fix the dockerize to properly compile using the monorepo

# Description
The dockerize was not properly using the new monorepo, but instead still building using the old package-lock which contained the old repo's for common. Now we are dockerizing on the root level, but only copying the packages that we need. **Please note that if we want to use more internal libraries, we need to also add each one seperately in the Dockerfile**

Fixes #214 

## Types of changes


apps/dashboard/Dockerfile
```diff
@@ -1,13 +1,15 @@
 # Build in a different image to keep the target image clean
-FROM node:20 as build
```
JustSamuel, line 2: :(

JustSamuel (approved):
20 en dan is priem

CodeNamedRobin (changes requested):
Pls better node version thnx

CodeNamedRobin (approved):
LGTM!

PR #224, opened 2024-06-21
fix(pos-settings): revert mistake in last PR

JustSamuel:
woeps


SuperVK (approved):
"Maar zou wel ff checken of `bottom: 0` miss nog iets doet"

Issue #216, opened 2024-06-18
[Refactor]: Create FormInput and default Dialog components 


JustSamuel:
### What browsers are you seeing the problem on?

Other

### Is this happening on the SudoSOS point of sale or on the SudoSOS dashboard?

Dashboard

### Current behaviour

We are lacking a common components for general usage (such as FormInputs and Dialogs)

### Steps to reproduce

\

SuperVK:
Completed in #290

Issue #15, opened 2023-09-24
Form validation, code duplication


CodeNamedRobin:
Seeing that right now the code for validating forms is quite repetitive it might be a good idea to think about a way to solve this. A solution could be based on writing a vee-validation helper based on the swagger spec. This would also include separating all different schema's into their own files.

SuperVK:
Solved/migrated to #273

PR #231, opened 2024-06-23
Refactor(dashboard): forms

JustSamuel:
# Description
This PR provides an example and implementation of a new way to structure form creation.

#216 
#15 

## Types of changes


JustSamuel:
@SuperVK @CodeNamedRobin This PR is also for discussion.

Any suggestions on moving the submission code to the form component? I think we can also generalize the creating of toasts.

SuperVK (approved):
I see the benefit in it, I would like to see an issue that tracks all the places where this is needed, otherwise we get even more random different implementations.

SuperVK (approved):
I see the benefit in it, I would like to see an issue that tracks all the places where this is needed, otherwise we get even more random different implementations.

SuperVK:
Thank you Github

PR #235, opened 2024-06-24
Feature(point-of-sale): AB testing

JustSamuel:
# Description
This PR adds an implementation of AB testing using a separate GitHub action. The idea of the action is that it builds the development branch with the production env variables such that it can be used next to the main branch side by side. 

This is nice for the point-of-sale as our develop branch will automagically become a canary release.

You can also view the tag and branch that is running in the bottom right of the login screen:
image

## Types of changes

### DIFF
--- a/apps/point-of-sale/env.d.ts
+++ b/apps/point-of-sale/env.d.ts
@@ -1,1 +1,3 @@
 /// <reference types="vite/client" />
+declare const __GIT_BRANCH__: string;
+declare const __GIT_COMMIT__: string;
--- a/apps/point-of-sale/src/components/GitInfo.vue
+++ b/apps/point-of-sale/src/components/GitInfo.vue
@@ -3,8 +3,8 @@
 </template>
 
 <script setup lang="ts">
-const branch = __VITE_GIT_BRANCH__;
-const commit = __VITE_GIT_COMMIT__;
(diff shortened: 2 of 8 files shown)


apps/point-of-sale/env.d.ts
```diff
@@ -1 +1,3 @@
 /// <reference types="vite/client" />
+declare const __GIT_BRANCH__: string;
+declare const __GIT_COMMIT__: string;
```
SuperVK, line 3: should be __VITE_GIT_XXX_ if im not mistaken

apps/point-of-sale/src/components/GitInfo.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <div class="absolute tag font-bold opacity-30"> {{branch}}#{{commit}} </div>
+</template>
+
+<script setup lang="ts">
+const branch = __GIT_BRANCH__;
+const commit = __GIT_COMMIT__;
```
SuperVK, line 7: idem

SuperVK (changes requested):
The tagging note does not render, I think (because I quickly testing, wont be schijnheilig here) because of inconsistent variables :)

SuperVK (approved):
Approve no scope als het nu niet werkt is t jouw schuld

PR #236, opened 2024-06-25
CI: semver and releases

JustSamuel:
# Description
The goal of this PR is to create automagic versioned releases of the monorepo and its libraries. They will be released in the form of `{{app}}:{{monorepo version}}`. The releases will be atomically done when merging into main based on `sementic-release` commits analyzer:

(Squashed) PRs with commits containing:
- `fix`: will cause a patch release
- `feat`: will cause a feature release (minor)
- `BREAKING CHANGE`: will cause a major release

none afaik.

## Types of changes


SuperVK:
If I read the yaml's correctly, the docker image ci run and push to pkg.gewis.nl on pushing to main. Now with the fast-forward this should not trigger, but if we do want to push it will trigger and still push to main tag on docker. Is this intended?

JustSamuel:
> If I read the yaml's correctly, the docker image ci run and push to pkg.gewis.nl on pushing to main. Now with the fast-forward this should not trigger, but if we do want to push it will trigger and still push to main tag on docker. Is this intended?

Yes, in this construction we will also have a rolling main, develop, sudopos-ab (which is develop)

You can decide to not do it, but since not every push releases a version it might be nice to have a "always latest of the branch". I.e someone did a wrong commit (ci but actually contained a bugfix( which can then be instant deployed if needed.

.releaserc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "branches": ["feature/semver"],
```
SuperVK: klopt dit? of is dit nog van testen?

JustSamuel: Foutje, kan jij dat naar 'main' omzetten?

.github/workflows/docker.yml
rinkp, line 1: Please note that you are renaming an action that is mandatory for PRs so upon merging this, please update the branch protection rules accordingly

rinkp (commented):
I have not tested the GH actions, but i am in favor of this idea.

Please note that you are renaming an action that is mandatory for PRs so upon merging this, please update the branch protection rules accordingly

SuperVK (approved):
lets test it!

PR #249, opened 2024-06-29
ci: add portainer webhook to rebuild stack

JustSamuel:
# Description
Automatically redeploy the dev stack when the images are updated. Makes developing more responsive. Once the backend is also versioned we could consider re-deploying the prod stack in the same manner such that sudopos-b will always be updated.

## Types of changes


SuperVK (approved):

BHenkemans:
image
:( de key is fout

PR #260, opened 2024-07-24
refactor(AdminUserBalance): Remove all multi-purpose related things

### LINKED ISSUE #258: [Bug]: Balance on `/user/[id]` page sometimes is current user's balance

Firefox, Microsoft Edge

Dashboard

### Current behaviour

**Sometimes**, when going to another user through `/user/[id]`, the balance of the signed-in user is shown instead of the `[id]` user's balance

image

### Desired behaviour

The expected behaviour is that the `[id]` user's balance is shown on their page

### Steps to reproduce

1. Go to `/user-overview`
2. Search for a user
3. Open the user

### Other information

What _bacpm_ and I think is happening is that one of two things happens:
1. There is a race condition, and sometimes the balance is replaced the one way, and sometimes the balance is replaced the other way
2. Something goes wrong with fetching the `[id]` user's balance and it is not shown. However, the network request was successful even when the issue occurred.

## Known workaround
I have showed the bacpm to view the network requests to see the balance. Alternatively, a CTRL+F5 works.

SuperVK:
Refactoring AdminUserBalance to remove all things related to this component being multi-purpose

# Description

This component is copy pasted from another component which was half multi-purpose, but also half only fetching the balance of the logged in user. This is now all removed, and this component is only used in one place, for users that are (most of the time) seperate from the logged in user. This fixes a few issues with regards to redundant options being passed, but also the race condition where it would show the logged in user's balance.

#258 

## Types of changes

### DIFF
--- a/apps/dashboard/src/modules/admin/components/AdminUserBalance.vue
+++ b/apps/dashboard/src/modules/admin/components/AdminUserBalance.vue
@@ -1,8 +1,6 @@
 <template>
   <CardComponent
       :header="$t('c_currentBalance.balance')"
-      :action="showOption ? $t('c_currentBalance.Increase balance') : ''"
-      routerLink="balance"
   >
     <div class="flex flex-column justify-content-center">
       <div v-if="isLoading">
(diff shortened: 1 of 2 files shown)


JustSamuel (approved):
lgtm

PR #262, opened 2024-07-24
Proxy api requests through vite

### LINKED ISSUE #256: Use Vite proxy to forward API requests and prevent CORS errors in development environment

Both

### What would you like?

Vite has the ability to forward all requests to a certain path to a different host. This can be used to prevent the CORS issues.

### Why is this needed?

The SudoSOS frontend development environment does not work out of the box due to these CORS errors. The developer is expected to disable CORS, which is a workaround (and a troublesome one) in my opinion and not a true fix.

### Other information

When this all works correctly, the production environment can also be configured such that CORS can be enforced.

SuperVK:
proxy api requests through vite to circumvent CORS issues

# Description

You do not need to disable CORS, because it is just making requests to the vite server, which will proxy it to the correct api.

Adds an extra env variable for the url to which you want to proxy, eg localhost:3000 or sudosos.test.gewis.nl/api/v1

#256 

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,3 @@ You can quickly start development directly in your browser by using [Codespaces]
 - Run `cd apps/dashboard` or `cd apps/point-of-sale` in the terminal. 
 - Run `npm run dev` to start the development environment.
 - You can access the dashboard and point of sale at `localhost:5173` and `localhost:5174` respectively.
-
-### Set up browser for CORS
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -66,6 +66,3 @@ You can quickly start development directly in your browser by using [Codespaces]
 - Run `cd apps/dashboard` or `cd apps/point-of-sale` in the terminal. 
 - Run `npm run dev` to start the development environment.
 - You can access the dashboard and point of sale at `localhost:5173` and `localhost:5174` respectively.
-
-### Set up browser for CORS
-By default no network requests are allowed from other hosts to `sudosos.test.gewis.nl`. This will prevent critical network requests like authentication from working correctly. To allow this, CORS will need to be disabled. This can be done by using an [extension](https://addons.mozilla.org/en-US/firefox/addon/cors-everywhere/) that can disable CORS. Alternatively, if you are using chrome, [you can directly run it without CORS](https://stackoverflow.com/questions/3102819/disable-same-origin-policy-in-chrome).
```
JustSamuel, line 71: Maybe explain how the CORS proxy works?

JustSamuel (approved):

PR #289, opened 2024-08-03
ci: fix webhook by ignoring result

JustSamuel:
# Description
Should pass even without getting a response

## Types of changes


BHenkemans (approved):
This seems like a reasonable solution. However, if, for whatever reason, something does actually go wrong, the build will still succeed. That said, the stack not restarting is not a massive issue.

PR #272, opened 2024-07-30
Refactor: form cards and utils

JustSamuel:
# Description
Its PR #231 but cooler. I added `FormInput`, which works best even though adding all the input types seems a bit weird. This solved some issues I had with the previous solution.

This is currently only used for the Invoices, but I made a tracking issue here: #273

## Related issues/external references
#216 
#15 

## Types of changes
- New feature _(non-breaking change which adds functionality)_

PR #277, opened 2024-07-31
fix!: client update

JustSamuel:
# Description
Updates to the latest client that is compatible with the dev backend.
issues/external references

## Types of changes
- Breaking change _(fix or feature that would cause existing functionality to change)_

PR #290, opened 2024-08-03
feat: viewing en editing invoices

JustSamuel:
# Description
This PR adds the general invoice and single invoice overview. The user is able to edit the invoice in the single invoice view.

The invoice overview is a group of 4 tables, each for every state:
image

The single invoice view is where most of the code and changes are. It is also based on the form refactor PR #272.

This is how a normal invoice looks like:
image

If the PDF is unavailable the user can switch to a tabular view using the `table` button:
image

The recompile and reload button do the same for now, since the backend still needs an update:
image

Pressing edit for one of the two cards will enable the user to edit the invoice fields:
image

Using the new form setup has the following benefits:
Closing the form resets it to defaults, not touching the form does not submit and it is super reactive.

The user can delete an invoice:
image

A deleted invoice wont show the edit options, and highlights that the invoice is deleted. In this case the PDF is also missing:
image

## Types of changes


lib/common/dist/sudosos-frontend-common.js
SuperVK, line 1: We should either rollback this or run build again, because right now it breaks the loading of mutations. 

#277 never included a building of the frontend-common, not sure why this specific version breaks, but building again works fine.

SuperVK:
Small quick remarks:

Would be nice to make the text based table overview more like the transaction overview, with right aligned totals and bold to differentiate.
image
image

Would be nice if VAT in the text based table has a percentage.

It seems that in the invoice overview the test invoice for the intro group shows 0 euro's, while the invoice itself has more?
image

apps/dashboard/src/modules/financial/views/invoice/InvoiceInfoView.vue
```diff
@@ -0,0 +1,63 @@
...
+    handleError(error, toast);
+  }).then((res) => {
+    invoice.value = res as InvoiceResponse;
+  });
+
+  if (!invoice.value) {
+    await router.replace('/error');
+    return;
+  }
+});
```
SuperVK, line 65: In this on purpose onBeforeMount, because now the loading is very weird on slower internet, where the footer is all the way at the top for half a second.

apps/dashboard/src/modules/financial/components/invoice/forms/InvoiceAddressingForm.vue
SuperVK, line 1: I am not fully sure about the placeholders, also because they are not i18n.

JustSamuel:
> Small quick remarks:
> 
> 
> 
> Would be nice to make the text based table overview more like the transaction overview, with right aligned totals and bold to differentiate.
> 
> image
> 
> image
> 
> 
> 
> Would be nice if VAT in the text based table has a percentage.
> 
> 
> 
> It seems that in the invoice overview the test invoice for the intro group shows 0 euro's, while the invoice itself has more?
> 
> image
> 
> 
> 
> 
> 
> 

It seems that in the invoice overview the test invoice for the intro group shows 0 euro's, while the invoice itself has more?

Invoice consists of custom amounts, so it shows 0 as the actual transaction in SudoSOS is 0 by design.

lib/common/dist/sudosos-frontend-common.js
JustSamuel: Maybe add a check to see if the build matches what is actually there?

SuperVK:
image
Do we want to keep this like this? or make this something more pretty?

JustSamuel:
> Do we want to keep this like this? or make this something more pretty?

fixed

SuperVK (approved):
Looks good enough for production!

PR #291, opened 2024-08-04
Feature: payout requests

JustSamuel:
# Description
This PR adds the creating and managing of `PayoutRequests` from the financial menu.

image

The main table contains a tab for each of the 4 states. The actions allow a user to view a `PayoutRequest` or download the PDF.
The user can inspect the payout by pressing the eye icon:

image

This inspection allows the user to deny or approve the payout. When first inspected, the specific information such as the users' balance and payment info still has to be loaded and is displayed as a skeleton.

image

After approving a payout, the table is reactively updated:

image

Non `CREATED` payouts will have a slightly different inspection dialog. 

image

An admin can also create a payout:

image

The user selection dropdown uses lazy loading

The form has validation (notice how the submit button is disabled), and also shows the users current balance:

image

Even a custom error when the balance of the user is lower than the requested payout:

image

After accepting, the form is reactively added to the table:

image

## TODO
- [x] translations
- [x] Refactor / clean up some code
- [x] Add loading indication when submitting form

## Types of changes

### DIFF
--- a/apps/dashboard/src/components/FindUser.vue
+++ b/apps/dashboard/src/components/FindUser.vue
@@ -19,15 +19,16 @@
 
 <script setup lang="ts">
 import { onMounted, type PropType, ref, watch } from "vue";
+import type { Ref } from "vue";
 import apiService from "@/services/ApiService";
 import { debounce } from "lodash";
-import type { UserResponse } from "@sudosos/sudosos-client";
+import type { BaseUserResponse, UserResponse } from "@sudosos/sudosos-client";
 
 const lastQuery = ref("");
 const selectedUser = ref(null);
 const loading = ref(false);
 
-const users = ref([]);
+const users: Ref<(BaseUserResponse & { fullName: string })[]> = ref([]);
 const emits = defineEmits(['update:value']);
 
 defineProps({
@@ -41,8 +42,8 @@ defineProps({
   },
 });
 
-const transformUsers = (userData) => {
-  return userData.map(user => ({
+const transformUsers = (userData: BaseUserResponse[]) => {
+  return userData.map((user: BaseUserResponse) => ({
     ...user,
     fullName: `${user.firstName} ${user.lastName}`
   }));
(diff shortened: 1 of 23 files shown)


SuperVK:
image
Might nice to add a hover tooltip for the right icon because it is not directly clear what it does. (I still don't know, because clicking it downloads a file without extension?)

image
Capatilize pls :upside_down_face: 

Creating doesn't reactively add to the created list, in this out of the scope of this PR?

Viewing a created payout removes it from the list. And I cannot approve it either.


I am not sure if I like the tab view inside the cardpanel, it's out of line with the invoices and the mutations view. I get it because you need the button for creating, but maybe there is another option?

JustSamuel:
> Viewing a created payout removes it from the list. And I cannot approve it either.

> Creating doesn't reactively add to the created list, in this out of the scope of this PR?

Fixed, backend dev still had to be redeployed.

> Capatilize pls 🙃

fixed

JustSamuel:
@SuperVK 

image

opinion?

JustSamuel:
> Might nice to add a hover tooltip for the right icon because it is not directly clear what it does. (I still don't know, because clicking it downloads a file without extension?)

It downloads the PDF, however the file is extensionless (should be fixable in the proxy)

image

SuperVK:
> opinion?

Yeah I think it's better, not ideal tho, the spacing seems a bit funky. Maybe a red button? We will have the same problem very shortly when the "Create invoices" will appear in your invoice overview

JustSamuel:
> > opinion?
> 
> Yeah I think it's better, not ideal tho, the spacing seems a bit funky. Maybe a red button? We will have the same problem very shortly when the "Create invoices" will appear in your invoice overview

image

you could hack it inline?

SuperVK:
The paginator is a bit broken, going back and forth doesn't really work, and the table is sorted ascending instead of descending (or the other way around, but not in the "normal" way atleast)

SuperVK:
Approving and denying payouts is also not reactively updated it seems

SuperVK:
> you could hack it inline?

I think that's the best option yeah

SuperVK (approved):

PR #295, opened 2024-08-10
CI: Add GitHub workflow for checking if build is up to date 

JustSamuel:
# Description
Simple CI check to make sure that the `dist` folder in lib/common is up to date.

## Types of changes


SuperVK:
Do we still want to do this? Or do we want to remove dist from the repo and dynamically integrate compiling common with the dev/build command for dashboard/pos?

JustSamuel:
> Do we still want to do this? Or do we want to remove dist from the repo and dynamically integrate compiling common with the dev/build command for dashboard/pos?

I think thats better.

PR #302, opened 2024-08-14
Feature/Refactor: seller and point of sale experience

### LINKED ISSUE #266: [Feature] Seller overview

Dashboard

### What would you like?

Possibly create a "seller overview" page where seller accounts can get some key info and also check some stats. Also allow them to create products within their own containers, because I have a feeling that the current product page does not like it when your not BACPM.

### Why is this needed?

Right now, only the BACPM can do these things, but ideally all sellers can do this. And it would be nice for various organs to see how much they sold during a borrel. 

### Other information

_No response_

SuperVK:
New interface for creating and managing point of sales. 

# Description

Hosts the point of sales under the seller tab instead of its own tab. Here you can find all the PoS that you have access to. (BACPM can see all). 

Per PoS you can see a quick overview, here you can edit basic information such as name, as well as edit the containers associated with the PoS. Either by directly creating one or by taking an existing container. 
image

Furthermore, there is a little overview of the sales in the last week, useful for especially borrels by both the BAC and other organs, to see how you sold in the last borrel. 

Fixes #266 

## Types of changes


apps/dashboard/src/modules/seller/components/POSOverviewTable.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <CardComponent :header="$t('app.Points of Sale')" class="w-full">
+    <DataTable :rows="rows" :value="pointOfSales" :rowsPerPageOptions="[5, 10, 25, 50, 100]" paginator lazy
+               @page="onPage($event)" :totalRecords="totalRecords">
+      <template #header>
+        <div class="flex flex-row align-items-center justify-content-end">
+
+          <Button :label="$t('app.Create')" icon="pi pi-plus" @click="openCreatePOSModal"/>
```
CodeNamedRobin: This should be in the same row as the header "Points of Sale" in the card. This can be done like in the point of sale info view in the overview card.

apps/dashboard/src/modules/seller/views/POSInfoView.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+  <div class="page-container flex flex-column">
+    <div class="page-title">{{ posName }}</div>
+    <div class="flex flex-column gap-5">
+      <div class="flex flex-column md:flex-row gap-5 justify-content-between align-items-stretch w-12">
+        <POSSettingsCard :pos-id="id!" class="flex-1 h-12" />
+        <CardComponent :header="$t('c_posInfo.lastWeekSales')" class="flex-1" >
```
CodeNamedRobin, line 7: Might be nice to also give this card h-12 class to line up with the POSSettingsCard.

apps/dashboard/src/modules/seller/views/POSInfoView.vue
```diff
@@ -0,0 +1,114 @@
...
+          <div class="h-12 text-center text-5xl pb-3">{{ formattedTotalSales }}</div>
+        </CardComponent>
+      </div>
+      <ContainerCard
```
CodeNamedRobin, line 11: There is no delete container from POS option :(

CodeNamedRobin, line 11: When clicking a product, no popup shows up. Background darkens, but modal is nowhere to be found.

CodeNamedRobin (changes requested):
Some small nitpicks.

apps/dashboard/src/modules/seller/views/POSInfoView.vue
SuperVK: The CSS doesn't CSS like that, I think I looked at it for about 15 minutes and then decided it was not worth my time. Could be fixed by modifying the CardComponent component, but wasn't worth it for this.

apps/dashboard/src/modules/seller/components/POSOverviewTable.vue
SuperVK: I was looking at this already, but couldn't found how to easily done it, but the topAction thing is what I need. Fixed!

apps/dashboard/src/modules/seller/views/POSInfoView.vue
SuperVK: Will fix in #303

SuperVK: Can't replicate? Might also be fixed in #309

apps/dashboard/src/components/InputContainerSpan.vue
```diff
@@ -0,0 +1,71 @@
...
+    default: false
+  },
+  column: {
+    type: Boolean,
+    required: false,
+    default: false
+  },
+});
+
+const container = defineModel<ContainerInStore>('container');
```
JustSamuel, line 65: Ahh, this is way nicer then how I did it for my UserInput, might have to look into this.

apps/dashboard/src/components/InputOrganSpan.vue
```diff
@@ -0,0 +1,68 @@
...
+      <span class="my-0">{{ label }}</span>
+      <Dropdown
+          v-model="organ"
+          :options="organs"
+          optionLabel="firstName"
+          :placeholder="placeholder"
+          class="w-full md:w-15rem"
+      >
+    <template #option="slotProps">
+      {{ slotProps.option.firstName + " " + slotProps.option.lastName }}
```
JustSamuel: NIT: extract to a function in the script body that also checks if there is a lastName?

```
function fullName(user: BaseUserResponse) {
   return user.lastName ? : user.FirstName + ' ' + user.Lastname : user.Firstname
}
```

apps/dashboard/src/main.ts
```diff
@@ -79,6 +78,13 @@ app.directive('tooltip', Tooltip);
 app.component('ToggleButton', ToggleButton);
 app.component('Steps', Steps);
 app.component('Calendar', Calendar);
+app.component('ConfirmDialog', ConfirmDialog);
 
-populateStoresFromToken(apiService);
-app.mount('#app');
+populateStoresFromToken(apiService).then(() => {
```
JustSamuel, line 83: Nice

JustSamuel (approved):

CodeNamedRobin (approved):

apps/dashboard/src/components/InputOrganSpan.vue
SuperVK: Done!

PR #303, opened 2024-08-14
feat: Permission based rendering

### LINKED ISSUE #265: [Refactor] Permissions based rendering per component

Dashboard

### What would you like?

We can use the new backend endpoint to render components based on permissions instead of based on roles. 

### Why is this needed?

Right now we do it per page, but one some pages people can see or do slightly more than other people. Furthermore, the roles we use right now are not fixed and might change, or dissappear. So we should do rendering based on the actual permissions that we know a user has, and that we can receive from the backend. 

### Other information

We could do it similiar to: https://github.com/GEWIS/parelpracht-client/blob/master/src/components/AuthorizationComponent.tsx and use a rolemanager similiar to the backend to determine what they can do.

SuperVK:
Render buttons and actions based on if they are actually allowed by the roles. See #265

# Description

Adds a helper function that can figure out what permissions somebody has. And applies to the product/containers, pos and the menu bar. 

Other views can be added later in the same way.

image
image

#265 

## Types of changes


apps/dashboard/src/modules/seller/components/ProductsCard.vue
```diff
@@ -31,28 +40,42 @@
...
+<!--              <i class="pi pi-upload"></i>-->
+<!--              <input-->
+<!--                  :ref="(el) => (fileInputs[rowDataImg.data.id] = el)"-->
+<!--                  type="file"-->
+<!--                  accept="image/*"-->
+<!--                  id="fileInput"-->
+<!--                  @change="(e) => onImgUpload(e, rowDataImg.data.id)"-->
+<!--              />-->
+<!--            </button>-->
+<!--          </div>-->
```
JustSamuel: ?

apps/dashboard/src/modules/seller/views/POSInfoView.vue
```diff
@@ -14,14 +15,18 @@
         :containers="posContainers"
         show-create
         :associatedPos="pointsOfSaleWithContainers[id!]"
+        :pos-edit-allowed="pointsOfSaleWithContainers[id!] &&
+        isAllowed('update', [getRelation(pointsOfSaleWithContainers[id!].owner!.id)], 'PointOfSale', ['any'])"
```
JustSamuel: I prefer to have limited logic in the html itself, I think for just pure premissions its fine, but this gets harder and harder to read (and to debug). Maybe turn this into a local scope helper function?

apps/dashboard/src/modules/seller/views/POSInfoView.vue
```diff
@@ -66,6 +72,11 @@ const posName = computed(() => {
   : t('common.loading');
 });
 
+const canLoadTransactions = computed(() => {
+  if (pointsOfSaleWithContainers.value[id.value!] == undefined) return false;
+  return isAllowed('get', [getRelation(pointsOfSaleWithContainers.value[id.value!].owner!.id)], 'Transaction', ['any']);
+});
+
```
JustSamuel, line 83: something like this as a helper function

JustSamuel (changes requested):
small nits

apps/dashboard/src/modules/seller/components/ProductsCard.vue
SuperVK: Thanks! removed it

apps/dashboard/src/modules/seller/views/POSInfoView.vue
SuperVK: Made it a helper function!

JustSamuel (approved):

PR #304, opened 2024-08-14
feat(dashboard/invoices): add ability to create invoices

### LINKED ISSUE #130: Implement overview of invoice accounts.
once this issue: https://github.com/GEWIS/sudosos-backend/pull/148 has been merged, we can implement these endpoints to be used by the frontend.

Basic idea is that you can see all of the invoice accounts that still have money open, you can generate invoices for these accounts, and later mark them as paid. As well deleting invoices etc.

CodeNamedRobin:
# Description
Adds invoice user table showing invoice users with outstanding debts, and allows for creating invoices. immediately.

## Types of changes


SuperVK:
image
Maybe I'm missing something, but this does not seem to work?

CodeNamedRobin:
> image Maybe I'm missing something, but this does not seem to work?

Hi Victor,
It should work, but is dependent on if invoice users have outstanding debts. In the latest commit I have added a string if there are no outstanding debts by invoice users. Should clear the confusion.

SuperVK:
I cannot see the string, there are translations missing as well. And I can't seem to link transactions to the invoice, or add some sort of amount to the invoice? 

Again, maybe I'm missing something here though. But then the UX might need to be a bit clearer?

CodeNamedRobin:
> I cannot see the string, there are translations missing as well. And I can't seem to link transactions to the invoice, or add some sort of amount to the invoice?
> 
> Again, maybe I'm missing something here though. But then the UX might need to be a bit clearer?

Woops, I did not push the commit, should work now.

apps/dashboard/src/utils/validation-schema.ts
```diff
@@ -99,4 +99,18 @@ export const addContainerObject =
...
+      addressee: yup.string().required(),
+      description: yup.string().required(),
+      date: yup.string().required(),
+      reference: yup.string().required(),
+      isCreditInvoice: yup.boolean().required().default(false),
+      street: yup.string().required(),
+      postalCode: yup.string().required(),
+      city: yup.string().required(),
+      country: yup.string().required(),
+      attention: yup.string().required(),
```
SuperVK: Attention is not required by the backend

CodeNamedRobin: Fixed in latest commit.

SuperVK:
The translations are still missing. 

I also at this point would advise against merging this with the state of the backend. We can't test the code because of the 501 it's throwing, and we expect some spec changes as well. It will just be broken code that needs to be rewritten either way. If you really want I won't stop you and we can still merge, but I think it will just be a very confusing lifeless components that are not much useful in this state.

SuperVK:
Closed on request of Robin

PR #313, opened 2024-08-26
Automatic common building and better dev enviroments

### LINKED ISSUE #296: [CI/CD] Automatic common building

Both

### What would you like?

It would be nice if the development of common is automatically started with the development command for dashboard/pos. And if the building of dashboard/pos also triggers a rebuilding of common.

### Why is this needed?

We can then remove the dist folder from the repo, and the dist and the source will not be different anymore

### Other information

_No response_

SuperVK:
Creates a better way to run the dev enviroments, as well as building

# Description

Encapsulates the common into the dashboard/pos commands (both dev and building), and allows starting from the root directory, instead of having to go in each folder. And the building now always first builds the local dependencies. So it is guaranteed that the typescript code from common is also what is compiled.

#296 

## Types of changes


apps/dashboard/src/modules/auth/views/AuthLocalView.vue
```diff
@@ -70,7 +70,7 @@ import { toTypedSchema } from '@vee-validate/yup';
 import * as yup from 'yup';
 import InputText from 'primevue/inputtext';
 import { useToast } from 'primevue/usetoast';
-import { AxiosError } from 'axios';
+import type { AxiosError } from 'axios';
```
JustSamuel, line 73: Outside the scope of this PR

JustSamuel (approved):

apps/dashboard/src/modules/auth/views/AuthLocalView.vue
SuperVK: yeah I thought yarn 4 introduced this a problem, because the builds failed, but I ended up fixing it by retrying it five times and it working. So should be not needed anymore indeed

PR #315, opened 2024-08-28
fix(fetchAllPages): Fetch pages does not take last page

### LINKED ISSUE #197: (transferred): fetchAllPages does not take the last page
see https://github.com/GEWIS/sudosos-frontend-common/issues/6

SuperVK:
Slight problem with fetchAllPages

# Description

#197 

## Types of changes


lib/common/src/helpers/PaginationHelper.ts
JustSamuel, line 1: So, I made the "take" an argument, however, I had the tendency to make it Number.MAX_SAFE_INTEGER everywhere. which was a mistake on my part. The backend ignores everything above 500, so I propose we drop the take from the argument of the fetchAllPages, and just use 500.

JustSamuel (approved):

lib/common/src/helpers/PaginationHelper.ts
SuperVK: yeah I dropped both arguments now, as the initial skip was never used and I can never ever see a use case where that is useful while still wanting all the other pages.

Issue #273, opened 2024-07-30
[Tracker] Form Refactoring


JustSamuel:
List of unrefactored forms:

- [x] `ContainerActionsDialog`
- [x] `ChangeInfoComponent`
- [x] `ProductActions`
- [x] `BannerDialog`
- [x] `AdminSingleUserView`
- [x] `AdminUserOverView`
- [x] `AuthLocalView`
- [x] `AuthResetView`
- [x] `FineView`
- [x] `POSCreateView`
- [x] `POSEditView`
- [x] `ChangePassword`
- [x] `ChangePin`
- [x] `BalanceWithTopup`

SuperVK:
POS done in #302

SuperVK:
ProductActions should receive high priority for refactor because there is a bug when editing products from organs that are not in your authStore and can thus not be prefilled in the schema (thus the form validation fails).

JustSamuel:
AuthLocalView done in #566

JustSamuel:
BannerDialog done in #567

JustSamuel:
Balance done in #578

Issue #268, opened 2024-07-26
[Feature] Create API key for other users


Yoronex:
### Is this a feature for the SudoSOS point of sale or for the SudoSOS dashboard?

Dashboard

### What would you like?

On a user's profile page, it would be nice if an admin can create an API key for that account. It is especially interesting for integrations with external services (like Aurora).

Jobbiesaus:
Closed by #455

PR #320, opened 2024-08-29
refactor(dashboard/user): profile view

CodeNamedRobin:
# Description
Refactors the entire profile view. Makes the profile view a lot neater, where no more cards are used for simple settings, instead all settings are in one user settings card.
Many of #273
## Types of changes


apps/dashboard/src/modules/user/views/UserProfileView.vue
```diff
@@ -1,51 +1,31 @@
...
-        <ChangePassword />
-      </div>
-      <div class="col-6 md:col-6">
-        <ChangeApiKey v-if="isAdmin"/>
-        <div class="flex flex-row mb-2 align-items-center">
-          <h3 class="mr-3">{{ t('profile.extensiveDataAnalysis') }}</h3>
-          <InputSwitch v-model="dataAnalysis" @update:modelValue="handleChange" />
-        </div>
-      </div>
+    <div class="flex md:flex-row flex-column-reverse justify-content-between gap-8">
```
SuperVK: I think gap-4 (or similiar) looks a lot better :)

SuperVK (commented):
It would be nice if the buttons for pop up change to a cursor mouse on hover, and I think you might as well change the password changing to be the same as the pin change? Not sure tbh, but this feels weird while they kinda do the same

And do we want to implement Roy's suggestion while we are at it: #268

SuperVK:
Also I can't comment on the code, but the profile view somehow fetches all users and uses that to get your GEWIS id. But this endpoint will soon become privileged so it might be nice to do this differently (besides it already being very cursed that we fetch _all_ users to just use one)

apps/dashboard/src/modules/user/views/UserProfileView.vue
CodeNamedRobin: Has been implemented!

CodeNamedRobin:
> Also I can't comment on the code, but the profile view somehow fetches all users and uses that to get your GEWIS id. But this endpoint will soon become privileged so it might be nice to do this differently (besides it already being very cursed that we fetch _all_ users to just use one)

A new get individual gewis user fetch function has been implemented in the user store now.

SuperVK (approved):

PR #324, opened 2024-09-02
refactor(dashboard/user-overview): refactored create user form

CodeNamedRobin:
# Description
Refactors the create user form on AdminSingleOverview to use the new form setup.

One of #273.

## Types of changes

### DIFF
--- a/apps/dashboard/src/modules/admin/components/users/forms/UserCreateForm.vue
+++ b/apps/dashboard/src/modules/admin/components/users/forms/UserCreateForm.vue
@@ -1,0 +1,111 @@
+<template>
+  <div class="flex flex-column gap-2">
+    <InputSpan :label="t('userDetails.First name')"
+               :value="form.model.firstName.value.value"
+               :attributes="form.model.firstName.attr.value"
+               @update:value="form.context.setFieldValue('firstName', $event)"
+               :errors="form.context.errors.value.firstName"
+               :disabled="!edit"
+               id="name" :placeholder="t('userDetails.First name')" type="text" />
+    <InputSpan :label="t('userDetails.Last name')"
+               :value="form.model.lastName?.value.value"
+               :attributes="form.model.lastName?.attr.value"
+               @update:value="form.context.setFieldValue('lastName', $event)"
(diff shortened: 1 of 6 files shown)


apps/dashboard/src/modules/admin/components/users/forms/UserCreateForm.vue
```diff
@@ -0,0 +1,111 @@
...
+
+setSubmit(props.form, props.form.context.handleSubmit(async (values) => {
+  const createUserRequest: CreateUserRequest = {
+    firstName: values.firstName,
+    lastName: values.lastName,
+    email: values.email || '',
+    nickname: values.nickname || '',
+    type: values.userType,
+    ofAge: values.ofAge,
+    canGoIntoDebt: values.canGoIntoDebt,
```
JustSamuel: Prefer object spreading

apps/dashboard/src/utils/validation-schema.ts
```diff
@@ -12,26 +12,23 @@ import type { ContainerInStore } from "@/stores/container.store";
...
         email: yup.string().email(),
         nickname: yup.string(),
-        userType: yup.mixed().required(),
-        isActive: yup.boolean().required().default(true),
+        userType: yup.number().required().default(4),
         ofAge: yup.boolean().required().default(false),
         canGoIntoDebt: yup.boolean().required().default(false),
-    })
-);
+    });
```
JustSamuel, line 24: Since we moved to modules, would it not make more sense to keep this definition closer to its usage? maybe the parent component which handles the forms? Just some food for thought

JustSamuel (approved):
NIT

apps/dashboard/src/utils/validation-schema.ts
CodeNamedRobin: Could be a .ts file in the module folder, I agree.

apps/dashboard/src/modules/admin/components/users/forms/UserCreateForm.vue
CodeNamedRobin: Fixed in latest commit.

PR #325, opened 2024-09-02
refactor(localisation): restructure the localisation files

### LINKED ISSUE #17: Rework translation files
Right now the translation files are quite messy and a bit messy, so these could be easily reworked.

CodeNamedRobin:
# Description
Restructures the localisation files into multiple, smaller bits. This makes it easier to manage all different translations.

Closes #17.

## Types of changes


SuperVK:
image
Is there no way to do this without the hacky way? Like properly have it at the first level

CodeNamedRobin:
> image Is there no way to do this without the hacky way? Like properly have it at the first level

Nope, there is no way to do this with the current ESLint plugin. And I do want to keep that plugin as it enforces that localisation actually works.

apps/dashboard/src/modules/admin/views/AdminSingleUserView.vue
```diff
@@ -1,22 +1,21 @@
 <template>
   <div class="page-container">
-    <div class="page-title">{{ `${currentUser ? currentUser.firstName : ''}'s profile` }}</div>
-
+    <div class="page-title">
+      {{ t('modules.admin.singleUser.profile', {user: currentUser ? currentUser.firstName : ''}) }}
```
Gijsdeman, line 4:
```suggestion
      {{ t('modules.admin.singleUser.profile', {user: currentUser ?? currentUser.firstName}) }}
```

apps/dashboard/src/locales/nl/components/forms.json
Gijsdeman, line 1: Is there a reason to keep the file even though it does not contain any translations?

apps/dashboard/src/locales/en/components/forms.json
Gijsdeman, line 1: Is there a reason to keep empty translation file?

.eslintrc.cjs
```diff
@@ -19,8 +21,8 @@ module.exports = {
         'vue/block-lang': ['error', { 'script': { 'lang': 'ts' } }],
         '@intlify/vue-i18n/no-raw-text': 'error',
         '@intlify/vue-i18n/no-missing-keys': 'error',
-        '@intlify/vue-i18n/no-missing-keys-in-other-locales': 'error',
-        '@intlify/vue-i18n/no-unused-keys': ['error', { 'enableFix': true, 'extensions': ['.js', '.vue', '.ts'] }],
+        '@intlify/vue-i18n/no-missing-keys-in-other-locales': ['error', ],
```
Gijsdeman:
```suggestion
        '@intlify/vue-i18n/no-missing-keys-in-other-locales': ['error'],
```

or even 

```suggestion
        '@intlify/vue-i18n/no-missing-keys-in-other-locales': ['error', {}],
```

apps/dashboard/src/components/container/ContainersCard.vue
```diff
@@ -1,12 +1,12 @@
...
+  <CardComponent :header="t('modules.seller.productContainers.containers.header')" class="p-0">
     <template #topAction>
       <div v-if="showCreate || associatedPos" class="flex justify-content-endg gap-2">
         <Button
             v-if="associatedPos"
             @click="openContainerAdd()"
-            outlined>{{ t('containersOverview.addExisting')}}</Button>
-        <Button @click="openContainerEdit()">{{t('app.Create')}}</Button>
+            outlined>{{ t('modules.seller.productContainers.containers.addExisting') }}</Button>
+        <Button @click="openContainerEdit()">{{t('common.create')}}</Button>
```
Gijsdeman: Maybe add something along the lines of [vue/mustache-interpolation-spacing](https://eslint.vuejs.org/rules/mustache-interpolation-spacing) to enforce consistent spacing for template syntax (most of the times you use single spacing around, in some rare cases you forget it).

apps/dashboard/src/utils/i18nUtils.ts
```diff
@@ -9,28 +27,65 @@ const i18n = createI18n({
     legacy: false,
     globalInjection: true,
     messages: {
```
Gijsdeman, line 29: Why are the `components` and `modules` fields present in the imported json files?

I think it my be clearer if you omit those fields from the imported json files (and thus move all fields up a level), and instead import the files as, for example, `footer_en_components`, and destruct those json files directly.

apps/dashboard/src/modules/admin/views/AdminSingleUserView.vue
CodeNamedRobin: This does not work as it then interprets the type of currentUser as `never`

apps/dashboard/src/locales/nl/components/forms.json
CodeNamedRobin: Not really, removed in latest commit.

apps/dashboard/src/locales/en/components/forms.json
CodeNamedRobin: Not really, removed in last commit.

.eslintrc.cjs
CodeNamedRobin: Resolved in latest commit.

apps/dashboard/src/components/container/ContainersCard.vue
CodeNamedRobin: Added, is indeed better.

apps/dashboard/src/utils/i18nUtils.ts
CodeNamedRobin: I don't know exactly what you mean? If you mean the keys in the json files themselves, that is so the lintr recognises everything. Other setups don't work, I've tried.

Gijsdeman (approved):

JustSamuel:
`yarn build` fails for me with 

```
error during build:
Could not resolve "../locales/en.json" from "src/utils/i18nUtils.ts"
```

PR #343, opened 2024-09-15
feat(pos/seller-selection): add outline buttons for cashiers

Yoronex:
# Description

Adds cashier roles to the screen with all POS owners. They are visually distinguishable by the buttons: POS owners have a solid button, while cashiers have an outlined button. This allows "Feuten" to also create transactions in their own name.

image

POS implementation of https://github.com/GEWIS/sudosos-backend/pull/222. Note that the dashboard implementation is still missing. However, this (probably) has less priority, as you only need to set the cashier roles once.

## Types of changes


CodeNamedRobin (approved):
LGTM

PR #344, opened 2024-09-15
fix(pos): balance not changing when locking buyer

Yoronex:
# Description

When locking in a user in borrel mode (for example an invoice account), the user's balance is never updated after each transaction. The manual fix is to reselect the user, but this PR will automatically update the user's balance after every transaction.

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/services/PointOfSaleSwitchService.ts
+++ b/apps/point-of-sale/src/services/PointOfSaleSwitchService.ts
@@ -25,7 +25,7 @@ export class PointOfSaleSwitchService {
 
       // Switch buyer back to user if authentication
       const cartStore = useCartStore();
-      cartStore.setBuyer(target.useAuthentication ? useAuthStore().getUser : null);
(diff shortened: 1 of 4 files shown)


CodeNamedRobin:
I don't know if it is this PR, but I'm getting a 400 on transactions, specifically when I'm on ABC werkmoment drinken POS, creating a transaction for me, created by a different person.

apps/point-of-sale/src/services/PointOfSaleSwitchService.ts
```diff
@@ -25,7 +25,7 @@ export class PointOfSaleSwitchService {
 
       // Switch buyer back to user if authentication
       const cartStore = useCartStore();
-      cartStore.setBuyer(target.useAuthentication ? useAuthStore().getUser : null);
+      cartStore.setBuyer(target.useAuthentication ? useAuthStore().getUser : null).then(() => {});
```
SuperVK, line 28: I can't immediately see what this is for?

SuperVK (approved):

PR #350, opened 2024-09-20
refactor(dashboard/banner-forms): refactored the banner forms to use new ones

CodeNamedRobin:
# Description
Makes sure to use the new forms in the banner components.

One of many #273 

## Types of changes


SuperVK:
Dates in the past are not properly handled, as this throws a 400 on the backend

Forms are not cleared when closing the dialog, so the image from one can be transferred to another if the other has no banner defined. As well as when clicking on create again

Would be nice if the InputNumber is suffixed with s to indicate the number is seconds not milliseconds

I am not sure if I agree with the placeholder name being "BAC Weekend" maybe just "Please insert banner name"

Form validation when not providing an end date does not work

JustSamuel:
Superseeded by #567

PR #393, opened 2024-10-18
fix: bug in common rootapi

JustSamuel:
# Description
Small bug in common which makes the rootapi unusable.

## Types of changes


CodeNamedRobin (approved):
Wie dit leest trekt een bak

Issue #351, opened 2024-09-23
[Feature] NFC Authentication


Yoronex:
### Is this a feature for the SudoSOS point of sale or for the SudoSOS dashboard?

Both

### What would you like?

Implementation of NFC Authentication on the frontend, which (preferably) includes an implementation in both the POS and the Dashboard. It contains the following functionality separated into standalone requirements. I have already made my personal prioritization, but feel free to change this.

### Must-have
- [x] Allow users to log in on the POS by reading an NFC chip. This should be directly on the PIN login page, or by opening a dialog on the PIN login page.
- [x] Allow users to set their NFC Authentication after logging in on the POS.
- [x] #572

### Should-have
- [x] Allow users to delete their NFC Authentication in the dashboard.

### Could-have
- [x] Allow selecting users to create a transaction for by scanning an NFC chip. For example, instead of typing in one's name to find them in SudoSOS, a BAC member can scan their NFC card during the borrel to find them immediately.

As far as I am aware, the could-have was not supported in SuSOS. However, it might be nice to include it as an extra possible identification flow. Furthermore, it could also be used to support a SudoSOS-card system (where you get a bronze/silver/gold/platinum NFC card based on your SudoSOS spending).

### Why is this needed?

People can then (finally) use their campus card or OV chipkaart to log in on the POS.

### Other information

Current implemented endpoints on the backend should support all aforementioned requirements. There is currently no way to find users based on their NFC Authenticator, but you can simply use the authentication endpoint to see if any user exists. The authentication endpoint also returns the user, and not only a JWT token.

JustSamuel:
> Current implemented endpoints on the backend should support all aforementioned requirements. There is currently no way to find users based on their NFC Authenticator, but you can simply use the authentication endpoint to see if any user exists. The authentication endpoint also returns the user, and not only a JWT token.

Returning a signed JWT token just to throw it a way seems iffy

SuperVK:
I think @noelvandevrie started on this on the working moment, with the idea that registering can also be done via phones, as they also have NFC scanners built in!

PR #421, opened 2024-10-19
Feature/nfc update

JustSamuel:
# Description
Adds NFC support to the dashboard and (in the future) the pos.

#351 

## Types of changes


apps/dashboard/src/modules/user/components/forms/ChangeNFCForm.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div class="flex flex-column justify-content-between gap-2">
+    <InputSpan :label="t('modules.user.profile.nfcNew')"
+               :value="form.model.nfcCode.value.value"
+               :attributes="form.model.nfcCode.attr.value"
+               @update:value="form.context.setFieldValue('nfcCode', $event)"
+               id="nfcCode" type="text"
```
JustSamuel: Implied "scan your NFC tag now" instead of a visible input box.

apps/dashboard/src/modules/user/components/forms/ChangeNFCForm.vue
```diff
@@ -0,0 +1,66 @@
...
+const props = defineProps({
+  form: {
+    type: Object as PropType<Form<yup.InferType<typeof editNFCSchema>>>,
+    required: true,
+  },
+  edit: {
+    type: Boolean,
+    required: false,
+    default: false,
+  },
```
JustSamuel: I think this can be removed

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -80,6 +87,58 @@ import { useConfirm } from "primevue/useconfirm";
...
+    if (!('NDEFReader' in window)) {
+      status.value = "Web NFC is not supported on this device.";
+      console.log('no');
+      toast.add({
+            severity: "error",
+            summary: "Unsupported",
+            detail: "Browser is not mobile and does not have Web NFC.",
+            life: 3000,
+          });
+    } else {
```
JustSamuel: This can be unnested into an if statement with a [early return](https://forum.freecodecamp.org/t/the-return-early-pattern-explained-with-javascript-examples/19364)

```ts
if (!('NDEFReader' in window)) {
  // error handling
  return;
}

// continue
```

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -80,6 +87,58 @@ import { useConfirm } from "primevue/useconfirm";
...
+      const ndef = new (window as any).NDEFReader();
```
JustSamuel: Is there  no typing declaration you can use? i.e. must this be an any?

This [answer ](https://stackoverflow.com/questions/60596876/ndefreader-in-typescript)suggest creating a global types extension to include the typing for the reader

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -80,6 +87,58 @@ import { useConfirm } from "primevue/useconfirm";
...
+      await ndef.scan();
+      toast.add({
+        severity: "info",
+        summary: "Scan started",
+        detail: "NFC Scan started, please hold your NFC card to your phone!",
+        life: 3000,
+      });
```
JustSamuel: Prefer to put the `success` toast in a `.then` to make sure its only shown if it actually works [read more about promises here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -80,6 +87,58 @@ import { useConfirm } from "primevue/useconfirm";
...
+        const tagId = event.serialNumber;
+        toast.add({
+          severity: "success",
+          summary: "Scan finished",
+          detail: "NFC Scan finished, NFC ID added.",
+          life: 3000,
+        });
+        console.log(tagId);
+        await apiService.user.updateUserNfc(useUserStore().getCurrentUser.user.id, { nfcCode: tagId });
+      };
```
JustSamuel: Try and use the `.then` here as well since you only want to show the 'NFC ID added` if the API call actually worked.

Note that we have a general API error helper which can be found [here](https://github.com/GEWIS/sudosos-frontend/blob/0be9f4395ce5e9be8a169f2a75aad75cc2775f5b/apps/dashboard/src/utils/errorUtils.ts#L17).

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -80,6 +87,58 @@ import { useConfirm } from "primevue/useconfirm";
...
+    }
+  } catch (error) {
+    toast.add({
+      severity: "error",
+      summary: "Scan finished",
+      detail: "NFC Scan finished, NFC ID added.",
+      life: 3000,
+    });
```
JustSamuel: Succes message in the catch block? Try and also use the Error helper here.

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
JustSamuel, line 1: Looks good, but it might be nicer to turn the actual NFC reading into its own component?
I suggest creating a `NFCReader` component which contains all the logic.

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
noelvandevrie: fixed

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
noelvandevrie: made new types directory with a typing file for the web-nfc related stuff.

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
noelvandevrie: done

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
noelvandevrie: done

apps/dashboard/src/modules/auth/views/AuthLoginView.vue
```diff
@@ -61,6 +61,7 @@ onBeforeMount(() => {
 const loginViaGEWIS = () => {
   window.location.href = `https://gewis.nl/token/${import.meta.env.VITE_APP_GEWIS_TOKEN}`;
 };
+
```
JustSamuel: ?

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -81,6 +88,54 @@ import { useToast } from "primevue/usetoast";
...
+  });
+
+  // Handle the NFC tag when it is detected
+  ndef.onreading = async (event: NDEFReadingEvent) => {
+
+    // Extract and log the NFC tag's ID (serial number)
+    const tagId = event.serialNumber;
+    console.log(tagId);
```
JustSamuel: `console.log`

apps/dashboard/src/modules/auth/views/AuthLoginView.vue
noelvandevrie, line 64: oeps

noelvandevrie: oeps removed added line

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
noelvandevrie: removed

PR #422, opened 2024-10-20
feat(dashboard): Add maintenance screen

RubenLWF:
# Description

This PR adds a maintenance screen that will be displayed when SudoSOS is under maintenance.
This information is fetched from the `rootApi.ping()` endpoint.

The maintenance screen will look like this:
image

## Types of changes


apps/dashboard/src/main.ts
```diff
@@ -84,11 +82,13 @@ app.component('Steps', Steps);
...
+await beforeLoad();
+
+try {
+  app.mount('#app');
+} catch {
+  clearTokenInStorage();
+  const authStore = useAuthStore();
+  authStore.logout();
+  app.mount('#app');
+}
```
SuperVK: Okay I think I hindsight I gave you the wrong suggestion in the working moment. I would make this:
```suggestion
beforeLoad().then(() => {
    app.mount('#app');
});
```
or 
```suggestion
await beforeLoad()
app.mount('#app');

```

apps/dashboard/src/utils/beforeLoadUtil.ts
```diff
@@ -0,0 +1,11 @@
+import { useSettingsStore } from '@/stores/settings.store';
+import { populateStoresFromToken } from '@sudosos/sudosos-frontend-common';
+import apiService from '@/services/ApiService';
+
+export default async function beforeLoad() {
+
+  const settingsStore = useSettingsStore();
+  await settingsStore.fetchMaintenanceMode();
+
+  await populateStoresFromToken(apiService);
```
SuperVK: And here catch the possible error from `populateStoresFromToken` and log out if need:
```js
clearTokenInStorage();
authStore.logout();
```

SuperVK (changes requested):

SuperVK (approved):

JustSamuel:
Do we also want to include the logic for admin bypass here? Or just keep that at the API for now

apps/dashboard/src/main.ts
```diff
@@ -84,11 +81,6 @@ app.component('Steps', Steps);
...
-    app.mount('#app');
-}).catch(() => {
-    clearTokenInStorage();
-    const authStore = useAuthStore();
-    authStore.logout();
-    app.mount('#app');
-});
+await beforeLoad();
+
+app.mount('#app');
```
JustSamuel: Top level await is not allowed, use a `.then` instead.

(this is also why you have a build error)

@SuperVK

JustSamuel (changes requested):

JustSamuel (approved):

RubenLWF:
Does something similar also need to be made for the POS?

I heard @SuperVK mention something about that a websocket first needs to be implemented in the backend

JustSamuel:
Kinda, the POS is not very often refreshed, so my idea was to have a websocket which indicates if the backend is up.
If you feel like taking a deeper dive in the backend, I have a branch which has an almost working version 👀

PR #434, opened 2024-10-30
feat: Endpoint for public keys

Jobbiesaus:
# Description
The tokens for stripe and GEWIS are now received from the backend. Also almost all env variables are now coded directly into the files.

Lastly I updated the README to make sure developers can still find the location of these variables.

## Types of changes

- New feature

### DIFF
--- a/apps/dashboard/vite.config.ts
+++ b/apps/dashboard/vite.config.ts
@@ -1,7 +1,9 @@
 import { fileURLToPath, URL } from 'node:url';
-
 import { defineConfig, loadEnv } from 'vite';
 import vue from '@vitejs/plugin-vue';
+
+const PROXY_URL = 'https://sudosos.test.gewis.nl';
+
 // https://vitejs.dev/config/
 export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd());
@@ -28,10 +30,15 @@ export default defineConfig(({ mode }) => {
     server: {
       port: 5173,
       proxy: {
-        '/api': {
-          target: env.VITE_DEV_API_BASE,
+        '/api/v1': {
+          target: PROXY_URL + '/api/v1',
           changeOrigin: true,
-          rewrite: (path) => path.replace(/^\/api/, ''),
+          rewrite: (path) => path.replace(/^\/api\/v1/, ''),
+        },
+        'static': {
+          target:  PROXY_URL + '/static',
+          changeOrigin: true,
+          rewrite: (path) => path.replace(/^\/static/, ''),
         }
       }
     }
--- a/apps/point-of-sale/src/services/ApiService.ts
+++ b/apps/point-of-sale/src/services/ApiService.ts
@@ -1,6 +1,6 @@
 import { ApiService } from '@sudosos/sudosos-frontend-common';
 
 const apiService = new ApiService(
-  import.meta.env.VITE_APP_API_BASE
+  window.location.origin + "/api/v1"
 );
 export default apiService;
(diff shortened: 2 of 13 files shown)


apps/dashboard/vite.config.ts
```diff
@@ -28,10 +28,15 @@ export default defineConfig(({ mode }) => {
...
+        '/api/v1': {
+          target: 'https://sudosos.test.gewis.nl/api/v1',
           changeOrigin: true,
-          rewrite: (path) => path.replace(/^\/api/, ''),
+          rewrite: (path) => path.replace(/^\/api\/v1/, ''),
+        },
+        'static': {
+          target: 'https://sudosos.test.gewis.nl/static',
+          changeOrigin: true,
+          rewrite: (path) => path.replace(/^\/static/, ''),
```
JustSamuel: Maybe change the sudosos.test.gewis.nl to a CONST at the top of the file? less work to only have to change a singel var.

apps/point-of-sale/vite.config.ts
```diff
@@ -27,8 +27,8 @@ export default defineConfig(({ mode }) => {
     server: {
       port: 5174,
       proxy: {
-        '/api': {
-          target: env.VITE_DEV_API_BASE,
+        '/api/v1': {
+          target: 'https://sudosos.test.gewis.nl/api/v1',
```
JustSamuel: same for consitentcy

JustSamuel (approved):

apps/point-of-sale/src/services/ApiService.ts
```diff
@@ -1,6 +1,6 @@
 import { ApiService } from '@sudosos/sudosos-frontend-common';
 
 const apiService = new ApiService(
-  import.meta.env.VITE_APP_API_BASE
+  window.location.origin + "/api/v1"
```
tomudding, line 4: This will not work in the POS, right? As it will result in `sudopos(.test).gewis.nl/api/v1` and not `sudosos(.test).gewis.nl/api/v1`?

apps/point-of-sale/src/utils/imageUtils.ts
```diff
@@ -4,8 +4,8 @@ export function getProductImageSrc(product: ProductResponse): string {
   if (!product.image) {
     return 'https://imgur.com/CS0aauU.png';
   }
-  return `${import.meta.env.VITE_APP_IMAGE_BASE}products/${product.image}`;
+  return `${window.location.origin}/static/products/${product.image}`;
 }
 export function getBannerImageSrc(banner: BannerResponse): string {
-  return `${import.meta.env.VITE_APP_IMAGE_BASE}banners/${banner.image}`;
+  return `${window.location.origin}/static/banners/${banner.image}`;
```
tomudding, line 10: Same here

apps/point-of-sale/src/services/ApiService.ts
SuperVK: yes, I had one neuron firing in my brain while helping Job that also realised this, I tactically ignored it. I think the cleanest is to fix this is the nginx routing? Otherwise we are guaranteed to be back where we started. But we will have to maintain two nginx routing then

tomudding: Yes, fixing it via NGINX would be the easiest. However, I do not know how well it will play with CORS?

JustSamuel: cors isn't about subdomains right?

JustSamuel: Try in test I suppose

PR #436, opened 2024-10-30
fix(dashboard/user-topup-modal): prevent clicking payment button multiple times

### LINKED ISSUE #435: [Bug]: "PAY" button can be pressed multiple times leading to multiple Payment Intent confirmations

Chrome, Firefox, Microsoft Edge

Dashboard

### Current behaviour

Accidentally pressing "PAY" multiple times may lead to multiple confirmations on Stripe's side of the top-up processing:
image

This results in a second `.requires_action` to be generated with another payment redirect link. Because of this, the original payment redirect link is no longer monitored and the payment is not accepted (and eventually refunded to the user).

### Desired behaviour

The "PAY" button is disabled after clicking it till the redirect (or Stripe fails).

### Steps to reproduce

1. Start a top-up with a valid amount
2. Select 'iDEAL' as payment processor
3. Accidentally press the "PAY" button twice (or introduce some artificial network delay / use touch input)

### Other information

ABC-2410-577

tomudding:
# Description

Pressing the "PAY" button starts the confirmation of the Payment Intent of the current top-up. This sends a request to Stripe which will (if successful) return a payment redirect link for iDEAL transactions.

Unfortunately, due to various reasons it is possible to accidentally press the "PAY" button twice. This can be due to user error, device issues, or network issues. The result is that sometimes, the request to confirm the Payment Intent is send twice. This means that two payment redirect links are generated by Stripe, but only the first one will be used.

Because Stripe has generated the second link (or even more) only that link is still active for Stripe. In other words, successful payments through the first link are not actively confirmed by Stripe. This means that that money is "temporarily" considered lost and the top-up is not successful.

The money will be returned to the user by Stripe within 5 business days, however, we should do everything in our power to ensure that this does not happen (especially because people sometimes need to do large top-ups).

This change ensures that the "PAY" button is disabled after pressing it until the user is redirected to the iDEAL payment page. I would have loved to use an `Idempotency-Key` instead, however, it appears that this is not directly supported by the `confirmPayment` method.

Fixes GH-435 and is directly related to ABC-2410-577.

## Types of changes

### DIFF
--- a/apps/dashboard/src/locales/en/common/common.json
+++ b/apps/dashboard/src/locales/en/common/common.json
@@ -106,6 +106,9 @@
                     "verify": "Verification Failed",
                     "details": "Payout amount does not match the amount of purchases. {amount} vs. {totalInclVat}",
                     "unable": "Unable to verify the payout."
+                },
+                "topUp": {
+                    "unable": "Unable to complete the top-up. Please try again later."
                 }
             }
         },
--- a/apps/dashboard/src/locales/nl/common/common.json
+++ b/apps/dashboard/src/locales/nl/common/common.json
@@ -106,6 +106,9 @@
                     "verify": "Verificatie mislukt",
                     "details": "Uitbetaling heeft een ongeldige bedrag. {amount} vs. {totalInclVat}",
                     "unable": "Kan de uitbetaling niet verifiëren."
+                },
+                "topUp": {
+                    "unable": "Het is niet gelukt om op te waarderen. Probeer het later opnieuw."
                 }
             }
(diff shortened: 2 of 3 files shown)


JustSamuel (approved):

tomudding:
I do not know the inner workings of Vue and browsers or their timings. As such, I added the early return to make it incredibly hard, if not impossible, for a user to create two confirmation requests by accident.

SuperVK (approved):
Code looks good! Dont have time to test, but if needed I can quickly do that

PR #438, opened 2024-11-02
fix: remove topup warning on landing page

JustSamuel:
# Description
After @tomudding his fix for #435 with PR #436 I think its fine to remove the warning. I also thought about changing the text, but people don't read the message anymore since it has been there for quite a while. Removing it will be the clearest message, and will most likely entice people to let us known if it still fails.

## Types of changes

### DIFF
--- a/apps/dashboard/src/modules/user/views/UserLandingView.vue
+++ b/apps/dashboard/src/modules/user/views/UserLandingView.vue
@@ -2,14 +2,6 @@
   <div class="page-container flex flex-column">
     <div class="page-title">{{ t('modules.user.landing.title') }}</div>
     <div class="content-wrapper gap-5 flex md:flex-column flex-column">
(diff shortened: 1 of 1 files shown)


SuperVK:
My original idea to change that v-if=true to v-if=false, then if we have another message at some point we can easily change the text, and don't have to reinvent the wheel (ideally we incorporate this in some backend feature that we can easily switch).

JustSamuel:
> My original idea to change that v-if=true to v-if=false, then if we have another message at some point we can easily change the text, and don't have to reinvent the wheel (ideally we incorporate this in some backend feature that we can easily switch).

I guess, but that is more clutter for something that can be very easily resolved. I suggest we do a general messaging motd system in the backend that we display.

SuperVK (approved):
yeah fair, the code is that complex so its not hard to bring back

PR #439, opened 2024-11-02
Feature/nfc login

noelvandevrie:
# Description
Ability to log in to the POS by using NFC tag id's, solves not being able to do this.
No changes made to the UI, it runs in the background.

#351 

## Types of changes


JustSamuel (approved):

SuperVK:
Is it correct that this PR removed the EAN login? Was this intended?

noelvandevrie:
> Is it correct that this PR removed the EAN login? Was this intended?

The EAN login function still exists in the LoginView, the component for it was just renamed to ScannersLoginComponent as we now have 2 scanners, it made more sense imo to put both in one file. If you look at how we implemented which scanner should be called I think it should clarify.

PR #440, opened 2024-11-04
feat: nginx config for pos

JustSamuel:
# Description
Modifies the NGINX config such that it all the services fall under the same config.

## Types of changes

### DIFF
--- a/apps/point-of-sale/vite.config.ts
+++ b/apps/point-of-sale/vite.config.ts
@@ -9,6 +9,7 @@ export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd());
 
   return {
+    base: "./",
     plugins: [
       vue(),
     ],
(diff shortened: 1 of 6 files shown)


apps/point-of-sale/vite.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd());
 
   return {
+    base: '/pos/',
```
SuperVK: How do you want to deal with A/B testing like /pos-beta/ etc?

SuperVK (approved):

apps/point-of-sale/vite.config.ts
JustSamuel: Yeah, its not yet ready. This is still an issue I am trying to fix.

My idea is to rewrite the paths using nginx and just always have the project routed at "./" which kinda works? but it doesn't seem to actually run.

PR #444, opened 2024-11-08
Chore: update main from develop

JustSamuel:
# Description
This release will contain some minor bug fixes here and there as well as two larger changes: 
- #439 will allow users to login with their nfc card.
- #440 changes how the sudosos stack is structured, moving everything to a single nginx container as entry point. This menas that the point of sale will now be served in a subpath. The main version will be served under `/pos/` and the develop under `/pos-develop/`

## Types of changes


SuperVK (approved):
We need to make sure sudopos.gewis.nl does not break before the CBC has the change to fix the URL that sudosos is loaded from.

JustSamuel:
Yes, this will need a more careful deployment.

PR #449, opened 2024-11-12
Fix stripe key being used as an object instead of a string.

JustSamuel:
# Description
The backend returns an object `{publicKey: value}` but the swagger says its just `value`. The spec should also be fixend in the backend.

## Types of changes


SuperVK (approved):
Classic prank from the backend

PR #453, opened 2024-11-12
Fix user being routed before app was ready

JustSamuel:
# Description
`router` was loaded before the app was ready, most importantly the tokens were not loaded / parsed, causing mistakes in the routing.

## Types of changes


SuperVK (approved):
Kinda odd that this used to work before? always won the race condition?

PR #455, opened 2024-11-13
Use store to get API key of user and admins can request to change

Jobbiesaus:
# Description
The change can be requested on others' pages in the user overview.

Admins can now change the users' API key from the user overview.

We now use the store of users to change the API key of users and Admins can do it on the users' overview.

Fixes#268 

## Types of changes


JustSamuel (approved):
Works on my machine

PR #456, opened 2024-11-14
feat(dashboard): button to directly go to user page in fine overview

### LINKED ISSUE #357: [Feature] Ability to click through from the fine overview

Dashboard

### What would you like?

It would be nice to be able to directly open a user's page from the fine overview, for checking more detail about their account. Similar to the user overview.

### Why is this needed?

This is needed because for example the BAC treasurer would like to check for each fine user if they are fully eligible or similar

### Other information

_No response_

Jobbiesaus:
No we can go directly from the fines overview to a user's page

Extra button in fine overview

# Description

BACPM can now immediately click on a button to move from the fines overview to the user's page.

Fixes #357 

## Types of changes


JustSamuel:
Would it be possible to add screenshots to the PR? Also nice for future PR's as it makes it easier to scan over them.

Jobbiesaus:
Capture

Now, if you click on the i it opens a new tab s.t. the BACPM does not lose any settings done on the original page.

JustSamuel:
Small nit, for branch names try and use kebab case. So `feature/click-through-from-fines` instead of `feature/clickThroughFromFines`

apps/dashboard/src/modules/admin/views/AdminUserOverView.vue
```diff
@@ -190,20 +190,36 @@ function debounce(func: (skip: number) => Promise<void>, delay: number): (skip:
...
+              !isActiveFilter.value,
+              ofAgeFilter.value,
+              undefined,
+              filters.value.type.value || undefined
+          )
+          .then((response) => {
+            totalRecords.value += (response.data._pagination.count || 0);
+            allUsers.value = allUsers.value.concat(response.data.records);
+          });
+    }
```
JustSamuel: What does this achieve? why the change?

apps/dashboard/src/modules/admin/views/AdminUserOverView.vue
```diff
@@ -246,7 +262,8 @@ async function handleInfoPush(userId: number) {
       (record) => record.id == userId
   );
   if (clickedUser) userStore.addUser(clickedUser);
-  router.push({ name: 'user', params: { userId } });
+  let route = router.resolve({ name: 'user', params: { userId } });
```
JustSamuel: prefer const

JustSamuel (changes requested):

apps/dashboard/src/modules/admin/views/AdminUserOverView.vue
Jobbiesaus: This was for another issue, I did an oepsie

apps/dashboard/src/modules/admin/views/AdminUserOverView.vue
Jobbiesaus: Will fix

JustSamuel (approved):

PR #457, opened 2024-11-14
fix: feedback-user-cannot-in-debt

SvenMokveldje:


apps/point-of-sale/src/components/Cart/CartActionsComponent.vue
```diff
@@ -1,4 +1,7 @@
 <template>
+  <Message v-show="stopCheckoutForUserInDebt" severity="warn">
+    You are a user that cannot go into debt!
+  </Message>
```
JustSamuel: I think a pop-up would be better, since now you can also click it "away" which is weird. 

In my opinion a better solution would be that it shows a pop-up when trying to checkout instead?

apps/point-of-sale/src/stores/cart.store.ts
```diff
@@ -20,10 +20,10 @@ export interface CartProduct {
 
 interface CartState {
   products: CartProduct[]
-  buyer: BaseUserResponse | null
+  buyer: UserResponse | null
   buyerBalance: DineroObjectResponse | null
   createdBy: BaseUserResponse | null
-  lockedIn: BaseUserResponse | null
+  lockedIn: UserResponse | null
```
JustSamuel, line 26: Not sure if this typing holds since it won't actually be a UserReponse right?

apps/point-of-sale/src/components/Cart/CartActionsComponent.vue
```diff
@@ -73,14 +84,15 @@ const finalizeCheckout = async () => {
...
+
   await cartStore.checkout();
   checkingOut.value = false;
   duration.value = 3;
   await logoutService();
 };
 const checkout = async () => {
   if (!enabled.value) return;
-
+  if (stopCheckoutForUserInDebt.value) return stopCheckout();
```
JustSamuel: Note that there is still an automatic checkout after 30 seconds.

If you have an order which you cannot place because you would go into debt, the automatic checkout will still try and place the order, which will fail and the session wont close.

JustSamuel (changes requested):

apps/point-of-sale/src/stores/cart.store.ts
SvenMokveldje: Well how are you going to get the canGoIntoDebt otherwise?

SvenMokveldje: I checked, it returns a full UserResponse not a BaseUserResponse

SvenMokveldje:
@JustSamuel I did now make it such that you first have to remove the items, this makes better code in my opinion as I otherwise would have to struggle and make ugly code with using multiple booleans when to open the dialog box.

This was it is also more responsive by immediately going away when removing items.

JustSamuel:
I think the modal kinda implies that you cannot remove any objects anymore, since usually everything is locked until the popup is closed.

apps/point-of-sale/src/stores/activity.store.ts
```diff
@@ -79,6 +79,7 @@ export const useActivityStore = defineStore('activity', {
 
     async onTimerEnd() {
       const cartStore = useCartStore();
+      if (cartStore.checkUserCannotGoIntoDebt()) return this.restartTimer();
```
JustSamuel: wasnt the idea to make this a logout?

apps/point-of-sale/src/stores/cart.store.ts
```diff
@@ -79,6 +79,16 @@ export const useCartStore = defineStore('cart', {
...
+    checkUserCannotGoIntoDebt(): boolean{
+      const buyer = this.buyer;
+      if (buyer) {
+        if (buyer.canGoIntoDebt) return false;
+        if (!this.buyerBalance) return false;
+
+        return (this.buyerBalance.amount - this.getTotalPrice) < 0;
+      }
+      return false;
+    },
```
JustSamuel: wait what does this do?

JustSamuel (changes requested):
small request

apps/point-of-sale/src/stores/cart.store.ts
SvenMokveldje: Changed the naming to: checkUnallowedUserInDebt should be a bit more clear

JustSamuel (approved):

PR #460, opened 2024-11-18
fix(dashboard):web-nfc

noelvandevrie:
# Description
Fix the unclear messaging provided in the toasts on the dashboard when trying to start scanning a NFC tag, and make the nfcCode sent to the backend consistent with how the POS sends it so the code scanned by web-NFC works on the POS, which it currently does not.

## Types of changes


apps/dashboard/src/locales/en/common/common.json
```diff
@@ -106,8 +106,10 @@
             },
             "error": {
                 "error": "Error",
-                "scanNotStarted": "NFC scan not started, please check permission regarding NFC.",
-                "unsupported": "Your browser is not mobile and does not have Web NFC."
+                "scanNotStarted": "NFC scan not started, please check permissions regarding NFC.",
+                "desktopUnsupported": "Browser is not mobile.",
```
SvenMokveldje: I would make this: NFC scan can only be done using a mobile phone.

That would make it more clear to the user what exactly the problem is

apps/dashboard/src/locales/nl/common/common.json
```diff
@@ -107,7 +107,9 @@
             "error": {
                 "error": "Error",
                 "scanNotStarted": "NFC scan niet gestart, kijk of NFC permissies kloppen.",
-                "unsupported": "Je browser is niet mobiel en heeft geen Web NFC."
+                "desktopUnsupported": "Browser is niet mobiel.",
```
SvenMokveldje: Same as the other comment.

SvenMokveldje (commented):
I added some comments, also rebase the branch as well.

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -89,16 +89,38 @@ import apiService from "@/services/ApiService";
...
+
+  if (!('NDEFReader' in window) && isIOS) {
+    toast.add({
+      severity: "error",
+      summary: t("common.toast.error.error"),
+      detail: t("common.toast.error.iosUnsupported"),
+      life: 3000,
+    });
+  }
+
```
JustSamuel: can be simplified by having a single branch to check if the NDEFReader is not in the window and then simply switch case what the detail should be.

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -93,32 +93,26 @@ async function startScan() {
...
+    else {
+      detail = t("common.toast.error.desktopUnsupported");
+    }
 
-  if (!('NDEFReader' in window) && isAndroid) {
     toast.add({
       severity: "error",
       summary: t("common.toast.error.error"),
-      detail: t("common.toast.error.androidUnsupported"),
+      detail: detail,
```
JustSamuel, line 112: `{detail}` can be used as a shorthand for  `{detail: detail}`

JustSamuel (approved):
small nit

apps/dashboard/src/modules/user/components/UserSettingsComponent.vue
```diff
@@ -114,7 +130,7 @@ async function startScan() {
   ndef.onreading = async (event: NDEFReadingEvent) => {
 
     // Extract and log the NFC tag's ID (serial number)
-    const tagId = event.serialNumber;
+    const tagId = event.serialNumber.toLowerCase().replace(/[^a-z0-9]/g, '');
 
     await apiService.user.updateUserNfc(props.user.id, { nfcCode: tagId })
         .then(() => {
```
JustSamuel: prefer naming tagId to nfcCode and using shorthand `{nfcCode}`

JustSamuel (approved):
mini nit

PR #463, opened 2024-11-23
chore: remove unused GitHub actions from POS

tomudding:
# Description

Came across these will looking into the `actions/upload-artifact@v3` deprecation. These were left behind after migrating from two repos to a monorepo and are not used.

## Types of changes


SvenMokveldje (approved):

JustSamuel:
thanks for checking! ❤️

PR #469, opened 2024-12-09
feat(dashboard): Redirect to previous URL after login

### LINKED ISSUE #452: [Feature] Redirect url after login

Dashboard

### What would you like?

When a user clicks on a link but gets redirected to the login, they should be redirected back to the location of the initial link.

### Why is this needed?

More professional, and later if we have more interesting pages on the website (statistics) it will enhance the user experience

### Other information

[This answer](https://stackoverflow.com/questions/45856929/redirect-to-requested-page-after-login-using-vue-router) on stack overflow explains nicely how to do this.

RubenLWF:
# Description

This PR ensures that if a user clicks on a link to a specific URL but needs to log in first, they will be redirected to the original URL after successfully logging in.

This is achieved by adding the page from the original link to the session storage before the user is redirected to the login page.
Then, after the user has logged in, the user will be redirected to the page in the session storage, or the home page if this is empty. After this it will be removed from session storage again.

I have decided to put this info in the session storage instead of in a query parameter as suggested in [the example from the issue](https://stackoverflow.com/questions/45856929/redirect-to-requested-page-after-login-using-vue-router). This because using it as a query parameter will not work for the "Login via GEWIS" button since this button redirects the user, losing the query parameters in the URL.

I have not been able to test this with the "Login via GEWIS" button since this logs me in on [sudosos.test.gewis.nl](https://sudosos.test.gewis.nl/) instead.

https://github.com/GEWIS/sudosos-frontend/issues/452

## Types of changes


SuperVK (approved):

RubenLWF:
Hi @SuperVK, 
I am still not completely sure how the routing functions. Do you know which case in `index.ts` is entered after the login through the GEWIS site?

SuperVK:
From the GEWIS website its gets back to a url with a token inside, then https://github.com/GEWIS/sudosos-frontend/blob/298207b7ee0910d7848961c855b932e2b62c005c/apps/dashboard/src/modules/auth/views/AuthLoginView.vue#L52 and https://github.com/GEWIS/sudosos-frontend/blob/298207b7ee0910d7848961c855b932e2b62c005c/apps/dashboard/src/modules/auth/views/AuthLocalView.vue#L101 make sure this token is traded in for an proper JWT token from SudoSOS. The first time you land on the auth you are not authenticated yet, so you land in the first if statement.

RubenLWF:
Great, thanks!

In that case I believe this should do the trick 🙂

JustSamuel (approved):
Looks good

PR #470, opened 2024-12-18
feat: Wrapped

SuperVK:
Kleintje voor de kerstgalaborel? Then #469 can actually be used 

# Description

image
Works best on mobile and in production

Code is ugly, but can be removed after December (or cleaned up and re-used next year?)

## Types of changes


JustSamuel (approved):
sick

PR #475, opened 2024-12-24
fix(dashboard): maintanance mode bg and handling

JustSamuel:
# Description
Was white, for some weird reason? not sure why the scss wasn't being loaded as expected.

This PR also makes sure that the maintenance view is loaded when the backend us not available.
image

## Types of changes

### DIFF
--- a/apps/dashboard/src/utils/beforeLoadUtil.ts
+++ b/apps/dashboard/src/utils/beforeLoadUtil.ts
@@ -5,8 +5,15 @@ import apiService from '@/services/ApiService';
 export default async function beforeLoad() {
 
   const settingsStore = useSettingsStore();
-  await settingsStore.fetchMaintenanceMode();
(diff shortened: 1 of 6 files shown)


apps/dashboard/src/utils/beforeLoadUtil.ts
```diff
@@ -5,8 +5,15 @@ import apiService from '@/services/ApiService';
 export default async function beforeLoad() {
 
   const settingsStore = useSettingsStore();
+  try {
   await settingsStore.fetchMaintenanceMode();
   await settingsStore.fetchKeys();
```
RubenLWF: A bit nit-picky but it might be nice to indent these since they are in a try block now

RubenLWF (approved):
Looks good to me 🚀 

(A check is failing 🙁)

PR #481, opened 2025-01-07
feat(dashboard/fines): add confirmation dialog for notification and handout

tomudding:
# Description

The BAC treasurer said she would like to confirm before sending fine warnings and handing out fines. This way, she can make sure that the right selection has been made (as it does require some filtering) before performing either action.

## Types of changes

### DIFF
--- a/apps/dashboard/src/locales/en/common/common.json
+++ b/apps/dashboard/src/locales/en/common/common.json
@@ -36,6 +36,7 @@
         "close": "Close",
         "cancel": "Cancel",
         "delete": "Delete",
+        "areYouSure": "Are you sure?",
         "owner": "Owner",
         "back": "Back",
         "na": "N/A",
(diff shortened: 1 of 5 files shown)


tomudding:
Other improvements I was thinking about (perhaps also for this PR?):
- also a progress spinner for handing out fines
- disable the button of the respective action after confirmation (so that you do not open the confirmation again - and more importantly, don't do the same thing twice)
- move var clearing in `notifyUsers` to a `finally` as in `handoutFines`

JustSamuel:
> * disable the button of the respective action after confirmation (so that you do not open the confirmation again - and more importantly, don't do the same thing twice)

I have been working on an "action" button which has most of these features, a nice spinner, and feedback on fail / pass. I'll look into using that for the fines.

JustSamuel (approved):

PR #511, opened 2025-02-26
feature(dashboard): add transaction id to header

noelvandevrie:
adds transaction id to header so the BACpm does not have to open mail to see the id

# Description
Adds transaction id to header.

## Types of changes

### DIFF
--- a/apps/dashboard/src/locales/en/components/mutations.json
+++ b/apps/dashboard/src/locales/en/components/mutations.json
@@ -14,7 +14,7 @@
       "user": "User transactions",
       "notYou": "Did you not make this transaction?",
       "modal": {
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/locales/en/components/mutations.json
JustSamuel, line 1: please also do this for the dutch translation

JustSamuel (changes requested):

apps/dashboard/src/locales/en/components/mutations.json
noelvandevrie: oke dan

JustSamuel (approved):

PR #525, opened 2025-04-10
fix(disable-text-selection): remove in keypadcomponent

noelvandevrie:
# Description
Disables text selection in the keypadcomponent in the point of sale.

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
+++ b/apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
@@ -47,6 +47,7 @@ const handleKeyClick = (key: string) => {
   height: var(--key-size);
   font-weight: bold;
   font-size: $font-size-larger;
+  user-select: none;
 
   &-row {
     gap: var(--key-gap-size);
@@ -67,6 +68,7 @@ const handleKeyClick = (key: string) => {
 
   > * {
     color: #d40000!important;
+    user-select: none;
   }
 }
 </style>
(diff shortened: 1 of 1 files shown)


apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
```diff
@@ -47,6 +47,7 @@ const handleKeyClick = (key: string) => {
   height: var(--key-size);
   font-weight: bold;
   font-size: $font-size-larger;
+  user-select: none;
```
Gijsdeman: Why not apply this on all children of `key` and remove it from the `outline` class (since that one is conditionally loaded either way)?

Gijsdeman (changes requested):

apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
noelvandevrie: so remove the one in `outline` and keep the one in `key`  ?

apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
```diff
@@ -67,6 +68,7 @@ const handleKeyClick = (key: string) => {
 
   > * {
     color: #d40000!important;
+    user-select: none;
```
Gijsdeman:
```suggestion

```

apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
Gijsdeman:
```suggestion
  > * {
    user-select: none;
  }
```

JustSamuel (approved):

PR #530, opened 2025-04-10
fix: add relative base to vite config

JustSamuel:


SuperVK (approved):
pretty sure we have tried this before and it broke some things, but let see

PR #531, opened 2025-04-11
refactor(eslint): switch to gewis eslint configs

### LINKED ISSUE #218: [Refactor] Prettier fix and enforce eslint/prettier
We should run a prettier such that code style is consistent, furthermore ideally prettier and eslint are first tried seperately in the ci/cd, and if they fail do not attempt other CI/CD checks, to prevent unnecessary usage.

SuperVK:
Refactor to switch to https://github.com/GEWIS/js-configs (atleast for prettier and eslint, tsconfig can be done later)

This PR is only for the config that changed, #532  is for the files that had to be changed to conform to eslint, #534 is where the prettier is run

# Description

## Types of changes


lib/common/eslint.config.js
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "eslint/config";
+// Very ironic, but without the extensions eslint does not work
+// eslint-disable-next-line import/extensions
+import root from "../../eslint.config.js";
```
Gijsdeman: See comment on `eslint.config.js` in root of project.

apps/dashboard/eslint.config.js
```diff
@@ -0,0 +1,50 @@
+import { defineConfig } from "eslint/config";
+import vueI18n from '@intlify/eslint-plugin-vue-i18n'
+import vueParser from "vue-eslint-parser"
+import * as typescriptParser from "@typescript-eslint/parser";
+// Very ironic, but without the extensions eslint does not work
+// eslint-disable-next-line import/extensions
+import root from "../../eslint.config.js";
```
Gijsdeman: See comment on `eslint.config.js` in root of project.

eslint.config.js
Gijsdeman, line 1: Perhaps consider removing this file and moving imports to each workspace instead of importing this config. This will clean up some weird-looking imports upstream.

Rules that you think are missing from the config(s) but should be applied to both workspaces can just be added to https://github.com/GEWIS/js-configs/tree/main/configs/eslint-config-vue instead, to slowly create a more consistent linting experience across projects.

JustSamuel (approved):

PR #533, opened 2025-04-12
feature(pos): Top Up Warning Modal

noelvandevrie:
# Description
Adds a TopUpWarningComponent containing a modal that urges the user to top-up their account balance if this is negative.
Screenshot 2025-04-12 at 13 47 09
Screenshot 2025-04-12 at 13 47 18

This problem hopefully solves high negative account balances of users, it should atleast encourage people to do so.
The Dialog has been put in to a seperate component so if decided in the future can be added to more places inside the point-of-sale.

## Types of changes


apps/point-of-sale/src/components/Keypad/KeypadComponent.vue
```diff
@@ -70,6 +70,7 @@ const handleKeyClick = (key: string) => {
 
   > * {
     color: #d40000!important;
+    user-select: none;
```
JustSamuel:
```suggestion

```

apps/point-of-sale/src/stores/cart.store.ts
```diff
@@ -85,6 +85,15 @@ export const useCartStore = defineStore('cart', {
         this.products.push(cartProduct);
       }
     },
+    checkBuyerInDebt(): boolean {
+      const buyer = this.buyer;
+      if (buyer) {
```
JustSamuel, line 90: early return?

JustSamuel (approved):
slight changes

apps/point-of-sale/src/stores/cart.store.ts
noelvandevrie: Fix all early returns of cartstore in other PR ?

PR #536, opened 2025-04-12
feat: use cookie based beta routing

JustSamuel:


SuperVK (approved):
I prefer sigma routing but ok

PR #537, opened 2025-04-13
Show green menu bar in beta

JustSamuel:
# Description
Show a green menu bar if the beta is enabled, I am aware that dropdown colors  dont work, but its just for a beta indicator.

image

## Types of changes

### DIFF
--- a/apps/dashboard/src/utils/betaUitl.ts
+++ b/apps/dashboard/src/utils/betaUitl.ts
@@ -1,0 +1,4 @@
+export function isBetaEnabled(): boolean {
+  const match = document.cookie.match(/(?:^|;\s*)X-Beta-Enabled=([^;]*)/);
+  return match?.[1] === 'true';
+}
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/utils/betaUitl.ts
```diff
@@ -0,0 +1,4 @@
+export function isBetaEnabled(): boolean {
+    const match = document.cookie.match(/(?:^|;\s*)X-Beta-Enabled=([^;]*)/);
```
SuperVK: Are you sure the cookie is not httpOnly?

SuperVK (approved):
Rather blue but ok ;)

apps/dashboard/src/utils/betaUitl.ts
JustSamuel: It is not

https://github.com/GEWIS/sudosos-frontend/blob/03239c1e989ab4ed7791cea22c14e11ae07ecc4d/lib/nginx/default.conf#L43

PR #540, opened 2025-04-14
feat(pos): green borrel splash for BAC

noelvandevrie:
# Description
Adds a green splash that switches when the POS is put into borrelmodus.
Screenshot 2025-04-14 at 11 14 38

## Types of changes


apps/point-of-sale/src/assets/borrel-splash.svg
JustSamuel, line 1: Prefer to use css to change the color of the css. Its the `fill` property, but you might need to touch up the svg.

JustSamuel (changes requested):

apps/point-of-sale/src/assets/borrel-splash.svg
noelvandevrie: ronde 2 dan

apps/point-of-sale/src/App.vue
```diff
@@ -1,26 +1,27 @@
...
+  opacity: 1;
+  pointer-events: none;
+  z-index: -10;
+  overflow: hidden;
+
+  position: fixed;
+  top: -25%;
+  right: 0;
+  height: 175%;
+}
```
JustSamuel: Can removed right?

JustSamuel (changes requested):

apps/point-of-sale/src/App.vue
noelvandevrie: done

JustSamuel (approved):

PR #543, opened 2025-04-19
feature(pos): maintenance mode with websocket

RubenLWF:
# Description

This PR adds a view called `MaintenanceView` that is layered on top of the regular POS application. If a message is received over the WebSocket that maintenance mode has been enabled, it becomes visible and prevents use of the POS.

The screen when maintenance mode is enabled looks like this:
image

https://github.com/GEWIS/sudosos-backend/pull/469

## Types of changes


RubenLWF:
See https://github.com/GEWIS/sudosos-frontend/pull/544

PR #544, opened 2025-04-19
Add websocket code in `common` and basic handling in POS

JustSamuel:
# Description
Using the new websockets in the backend by @RubenLWF we are now able to inform users realtime about their connection and `maintenanceMode`

The POS will now show a "lost connection" overlay if the websocket disconnects:
image

Socket.io will automatically try to restore the connection and the overlay will disappear. In that case, the pos will try to refresh the users token, since if that fails the token is most likely expired and the user will be redirected to the login screen.

If `maintenanceMode` gets enabled the POS will be show a info screen in real time:
image

The user is also logged out.

## Types of changes


lib/common/src/stores/websocket.store.ts
```diff
@@ -0,0 +1,22 @@
+import { defineStore } from 'pinia';
+
+export const useWebSocketStore = defineStore('websocket', {
+  state: () => ({
+    connected: false,
```
RubenLWF, line 5: Do we need to store both `connected` and `disconnected`?

lib/nginx/default.conf
RubenLWF, line 1: Might be good for someone else to also review this since I do not really understand nginx 🤷

RubenLWF (changes requested):
Nice work!

I think that it might also be nice to ping the backend on refresh (like we do in the dashboard). This so that if the POS is refreshed while it is in maintenance mode, the WS message does not have to be emitted again for the overlay to show.

JustSamuel:
> Nice work!
> 
> I think that it might also be nice to ping the backend on refresh (like we do in the dashboard). This so that if the POS is refreshed while it is in maintenance mode, the WS message does not have to be emitted again for the overlay to show.

My idea was to emit the maintenance-mode on connect with the websocket from the backend, as that would make the code cleaner imho, instead of having a seperate API call

lib/common/src/stores/websocket.store.ts
JustSamuel: It was to prevent "flickering" on load,. The difference is that disconnect is initially false, meaning its more like a falling edge detector.

RubenLWF:
> > Nice work!
> > I think that it might also be nice to ping the backend on refresh (like we do in the dashboard). This so that if the POS is refreshed while it is in maintenance mode, the WS message does not have to be emitted again for the overlay to show.
> 
> My idea was to emit the maintenance-mode on connect with the websocket from the backend, as that would make the code cleaner imho, instead of having a seperate API call

That does indeed seem better, good idea!

RubenLWF (approved):

PR #548, opened 2025-05-03
fix: show fine warning only when initially logging in as a user, not …

noelvandevrie:
…when being selected within the POS

# Description

Fixes the negative balance warning showing when already logged in to the pos and selecting an user.

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/views/CashierView.vue
+++ b/apps/point-of-sale/src/views/CashierView.vue
@@ -7,7 +7,11 @@
   <div v-else class="main-content">
     <div class="wrapper">
       <div class="pos-wrapper">
-        <TopUpWarningComponent></TopUpWarningComponent>
+        <TopUpWarningComponent
+          v-if="shouldShowTopUpWarning"
+          :show="showTopUpWarning"
+          @update:show="handleTopUpWarningUpdate"
+        />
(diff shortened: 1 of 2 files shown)


apps/point-of-sale/src/views/CashierView.vue
```diff
@@ -58,6 +62,22 @@ enum PointOfSaleState {
 
 const currentState = ref(PointOfSaleState.DISPLAY_POS);
 
+const showTopUpWarning = ref(false);
+const hasShownTopUpWarning = ref(false);
+
+const shouldShowTopUpWarning = computed(() => {
+  const inDebt = cartStore.checkBuyerInDebt();
+  const notShown = !hasShownTopUpWarning.value;
+  return inDebt && notShown;
```
RubenLWF: You can simplify this into the following:
```suggestion
  return inDebt && !hasShownTopUpWarning.value;

```

RubenLWF (commented):
I believe this should work, I left 1 small comment.

apps/point-of-sale/src/views/CashierView.vue
noelvandevrie: Simplified in last commit

JustSamuel (approved):

Issue #353, opened 2024-09-23
[Chore] Migrate to PrimeVue V4


CodeNamedRobin:
[PrimeVue v4](https://github.com/primefaces/primevue/releases/tag/4.0.0) has been released on the July 3rd, 2024, and SudoSOS should migrate to it from v3, using the [migration guide](https://primevue.org/guides/migration/v4/)
### Is this for POS or Dashboard
_Both_
### What must be done?
- [x] Refactor `Calendar` to `DatePicker`
- [x] Refactor `Dropdown` to `Select`
- [x] Refactor `InputSwitch` to `ToggleSwitch`
- [x] Refactor `TabMenu` to use Tabs without Panels
- [x] Refactor `Accordion` to use `AccordionHeader` and `AccordionContent`
- [x] Recreate custom themes for SudoSOS

SuperVK:
Ideally we make a new package that handles the theming, so this can be used by both the POS and the dashboard. Ideally we set this up in such a way that it allows with little changes:
- Create a dark theme for the dashboard
- Creating custom themes for the POS, such that we can easily switch from normal red, to BAC green, or blue/pink Athena, or BOOM black/lime-green, I.V.V. Bordeaux/yellow

SuperVK:
Dashboard is done, now POS is up

SuperVK:
POS is done!

PR #561, opened 2025-06-22
Add reactive conditional theme switching

JustSamuel:
# Description
Following the primevue v4 update #353 we now have cool and hip preset switching.

Specifically we can register presets using `useConditionalPreset`, which will then be watched and reactively applied.

Use cases, currently implemented:

image
Normal preset because user is logged out **and not in beta**.

User logs in and is part of a specific organ. This organs theme is then applied (dynamically)
 
image

User enables beta and receives the `beta blue` theme:

image

Note that themes have an "implicit" importance. From top to bottom, so the last "passing" condition will be applied.

## Types of changes

### DIFF
--- a/apps/dashboard/src/components/FormCard.vue
+++ b/apps/dashboard/src/components/FormCard.vue
@@ -1,7 +1,7 @@
 <template>
   <CardComponent :header="header">
     <template #topAction>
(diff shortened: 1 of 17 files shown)


noelvandevrie (commented):
looking very niceee

SuperVK (approved):

PR #563, opened 2025-06-22
chore: upgrade POS to primevue v4

SuperVK:
# Description
 Also upgrade the POS to primevue v4

## Types of changes


SuperVK:
Only thing that still needs to be fixed is the product flying. But in the mean time it can be checked!

JustSamuel (approved):

PR #564, opened 2025-06-23
Add dark mode toggle

JustSamuel:
# Description

Wow! Adds a composable for dealing with dark mode. Preference is stored in local storage under `dark-mode`.

image

image

This might  require some style changes as we go. But anyhow.

## Types of changes


SuperVK:
Maybe we can make this match the system setting? That is pretty common standard on the web nowadays

SuperVK (approved):

PR #566, opened 2025-06-23
fix: styling of local login

JustSamuel:
# Description
Old design:
image

New design:
image

Also extended and fixed some of the form input things. (such as now adding an inline-error prop.)

## Types of changes


SuperVK (approved):
Maybe directly include the password reset form? Otherwise we will forget about that for the rest of eternity probably

PR #567, opened 2025-06-24
chore: move banner to new form setup

JustSamuel:
# Description

Actually just a remake of #350 

image

image

Uses the new dialog components for a more streamlined look.

Improved form validation:
image

Dates cannot be in the past and duration must be greater than 0.

image

## Types of changes

### DIFF
--- a/apps/dashboard/src/components/InputSpan.vue
+++ b/apps/dashboard/src/components/InputSpan.vue
@@ -86,6 +86,14 @@
         type="password"
         v-bind="attrs"
       />
+      <InputNumber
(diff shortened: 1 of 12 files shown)


SuperVK:
Start date is allowed to be in the past according to [the spec](https://github.com/GEWIS/sudosos-backend/blob/develop/src/service/banner-service.ts#L65), useful to create banners that start right away (and edit ongoing banners)

1000 seconds is quite a lot as default, perhaps 5-10 seconds?

SuperVK (approved):

PR #573, opened 2025-06-24
Delete Containers from POS

### LINKED ISSUE #349: [FEATURE] Deleting containers from POS

Dashboard

### What would you like?

Delete containers from point of sales

### Why is this needed?

The BAC PM needs this.

### Other information

_No response_

JustSamuel:
# Description

New feature that allows user to remove container from POS. Also adds some style fixing and some deprecations. 

Old design:
image
image

New design:
image
image

New feature:
image
image

Closes #349 

## Types of changes


SuperVK:
~Maybe the close button can be outlined like the rest?~ nvm this is inconsistent elsewhere as well

JustSamuel:
> ~Maybe the close button can be outlined like the rest?~ nvm this is inconsistent elsewhere as well

Maybe we should just make our own "useConfirmDialog" because this one doesn't do the "go away if clicked away" and we could apply our own styling? I was thinking about it, but that isn't the scope of this PR. We could simply copy the API and it would be plug n play

JustSamuel:
> ~Maybe the close button can be outlined like the rest?~ nvm this is inconsistent elsewhere as well

I am also extremely against dialog on dialog.

SuperVK (approved):

PR #577, opened 2025-06-25
fix: styling of mutations table on phones

JustSamuel:
# Description
Smal QOL for phones. Added a composable to use tailwind CSS breakpoints as javascript vars.

Old:
WhatsApp Image 2025-06-25 at 23 40 03_9c1d1185

New:
WhatsApp Image 2025-06-25 at 23 40 03_40c7dc71

## Types of changes

### DIFF
--- a/apps/dashboard/src/composables/sizeBreakpoints.ts
+++ b/apps/dashboard/src/composables/sizeBreakpoints.ts
@@ -1,0 +1,28 @@
+import { ref, onMounted, onUnmounted, computed } from 'vue';
+
+const breakpoints = {
+  sm: 640,
+  md: 768,
+  lg: 1024,
+  xl: 1280,
+  '2xl': 1536,
+};
+
+export function useSizeBreakpoints() {
+  const width = ref(window.innerWidth);
+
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/composables/sizeBreakpoints.ts
```diff
@@ -0,0 +1,28 @@
+import { ref, onMounted, onUnmounted, computed } from 'vue';
+
+const breakpoints = {
+  sm: 640,
+  md: 768,
+  lg: 1024,
+  xl: 1280,
+  '2xl': 1536,
```
SuperVK, line 8: possible to match this to rems? not sure if that is possible even or if rems are actually used

SuperVK (approved):

apps/dashboard/src/composables/sizeBreakpoints.ts
JustSamuel: Sadly you only get the actual pixel width. You could do rem and x16 but that doesn't provide anything extra

PR #580, opened 2025-06-26
feat: add ATHENA color theme

JustSamuel:
# Description

Its pink.

WhatsApp Image 2025-06-26 at 03 05 41_ab09abe5

image

## Types of changes

### DIFF
--- a/apps/dashboard/src/composables/organMember.ts
+++ b/apps/dashboard/src/composables/organMember.ts
@@ -1,11 +1,11 @@
 import { useAuthStore } from '@sudosos/sudosos-frontend-common';
 import { computed } from 'vue';
 
-export function useOrganMember(organId: number) {
+export function useOrganMember(organName: string) {
   const authStore = useAuthStore();
 
   return computed(() => {
     const organs = authStore.organs ?? [];
-    return organs.some((organ) => organ?.id === organId);
+    return organs.some((organ) => organ?.firstName?.toUpperCase() === organName.toUpperCase());
   });
 }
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/composables/organMember.ts
```diff
@@ -1,11 +1,11 @@
...
 import { computed } from 'vue';
 
-export function useOrganMember(organId: number) {
+export function useOrganMember(organName: string) {
   const authStore = useAuthStore();
 
   return computed(() => {
     const organs = authStore.organs ?? [];
-    return organs.some((organ) => organ?.id === organId);
+    return organs.some((organ) => organ?.firstName?.toUpperCase() === organName.toUpperCase());
```
SuperVK, line 9: why this change?

SuperVK (approved):
really like the blue/pink complement, tried that also a bit in the examples I sent, but this implementation is a lot better!

apps/dashboard/src/composables/organMember.ts
JustSamuel: ORGAN Id is not consistent across test and prod. But maybe checking an Id is still better? What do you think

SuperVK: no I just was about to come back to say that it makes sense that code should not make assumptions on the state of the database in terms of IDs

JustSamuel: Preferably we store organ and theme info in the backend. But this will go with a "customization PR" which we might do in the future

PR #581, opened 2025-06-26
feat: add I.V.V color theme

JustSamuel:
# Description
image

image

## Types of changes

### DIFF
--- a/lib/themes/src/color-themes/ivv-navy.ts
+++ b/lib/themes/src/color-themes/ivv-navy.ts
@@ -1,0 +1,50 @@
+import { definePreset } from '@primeuix/themes';
+import { SudososPreset } from '../sudosos';
+
+export const IvvNavy = definePreset(SudososPreset, {
+  semantic: {
+    colorScheme: {
+      light: {
+        primary: {
+          color: '#14213d', // navy blue as primary
+          inverseColor: '#ffffff',
+          hoverColor: '#23325b', // slightly lighter navy for hover
+          activeColor: '#14213d',
+        },
+        highlight: {
+          background: 'rgba(0,0,0,0)',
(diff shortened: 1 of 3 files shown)


lib/themes/src/color-themes/ivv-navy.ts
```diff
@@ -0,0 +1,50 @@
+import { definePreset } from '@primeuix/themes';
+import { SudososPreset } from '../sudosos';
+
+export const IvvNavy = definePreset(SudososPreset, {
```
SuperVK, line 4: Look I'm not I.V.V, but isn't navy a blue color? and isn't this Bordeaux?

SuperVK (approved):
Together with the comments for the Athena PR

lib/themes/src/color-themes/ivv-navy.ts
SuperVK: Apparently Bordeaux is Dutch, but the translation is Wine Red, which I think is pretty funny
https://nl.wikipedia.org/wiki/Bordeauxrood <-> https://en.wikipedia.org/wiki/Wine_red
https://nl.wikipedia.org/wiki/Marineblauw <-> https://en.wikipedia.org/wiki/Navy_blue

SuperVK: Also now see that the buttons are navy, didn't notice that in the screenshots

SuperVK:
also what is the priority on the organs? Cause we have quite a few members of both I.V.V and BAC?

JustSamuel:
> also what is the priority on the organs? Cause we have quite a few members of both I.V.V and BAC?

Beta is most important, then fraternities, then BAC.

PR #582, opened 2025-06-26
feat: add preload composable and option to mutations balance

JustSamuel:
# Description

So the whole reason I refactored the mutation modal was to do educated prefetching.

Old:
Recording 2025-06-26 at 11 45 02

New:
Recording 2025-06-26 at 12 01 41

## Types of changes


SuperVK (approved):

SuperVK:
image
AFter looking a bit closer, it seems like every mutation is fetched twice?

Also does this not put unnecessary strain on the backend?

PR #583, opened 2025-06-26
refactor: navbar content as individual modules

JustSamuel:
# Description

Generale improvement. Navbar content is now moved to modules instead.

Design stayed the same:
image

Only slight changes to the mobile bar:
Old:
image

New:
image

## Types of changes


SuperVK (approved):
very clean approach, I like it!

PR #594, opened 2025-06-27
feat: add B.O.O.M. theme and a secondary styling color

JustSamuel:
# Description

Main bonus is the usage of `color.secondary` for dialog headers, card titles and tabs.

image
image

## Types of changes


apps/dashboard/src/assets/theme-overrides.css
SuperVK, line 1: This should be possible via the definePreset right?

For example like: https://github.com/primefaces/primeuix/blob/main/packages/themes/src/presets/aura/tabs/index.ts

SuperVK (changes requested):
Maybe look into getting the SudoSOS logo to respect dark themed top navbars?

apps/dashboard/src/assets/theme-overrides.css
JustSamuel: image

SuperVK: that does not answer my question

SuperVK: I'm saying you do not have to create a separate .css file for overwriting the styling. That should be done via de primevue presets api

SuperVK (approved):

PR #596, opened 2025-06-27
feat: add defi lilac theme

JustSamuel:
# Description

image
image

## Types of changes


SuperVK (commented):
I think the topnavbar in dark theme can be the original color if the text is black?
And Défi calls the color Lilac instead of Magenta. (which from looking at Wikipedia is indeed closer to their color).

JustSamuel:
> I think the topnavbar in dark theme can be the original color if the text is black? And Défi calls the color Lilac instead of Magenta. (which from looking at Wikipedia is indeed closer to their color).

Hahahah pffff

JustSamuel:
fixed
image

SuperVK (approved):

PR #620, opened 2025-09-23
fix: do not pre-fill balance field with 0

SuperVK:
# Description

When loading the page, the balance field is null, instead of 0. This means you can start typing right away, instead of having to actually remove the 0,00 that is there by default. very ergonomic very nice

## Types of changes


RubenLWF (approved):
LGTM! :rocket:

JustSamuel (approved):

PR #626, opened 2025-09-30
chore: bump sudosos-client in common

JustSamuel:
Updated the client to use a more general output to prevent type errors.


RubenLWF (approved):
LGTM :rocket:

PR #634, opened 2025-10-17
feat: add the polish language

SuperVK:
# Description
A small easter egg feature for the fact that we reached the milestone of 100k transactions during the Polish social drink. All generated using AI. We can remove this after a few months, as it probably isn't feasible to maintain a language we all do not speak. 

## Types of changes


JustSamuel (approved):
LGTM _(I do not speak polish)_

PR #645, opened 2025-10-25
Edit transactions overview

### LINKED ISSUE #121: Let BAC PM edit transactions
Right now we get a lot of requests from people saying their transaction was wrong. Most of the time the transaction is for the wrong person. This should be easily editable in the frontend, by using the PATCH endpoint on /transactions

JustSamuel:
# Description
Interface for transaction editing.

Modal now shows an _edit_ button.
image

This brings you to the new transaction edit view!
image

with the query pre filled in. You can also find this under admin/transactions:
image

Which will bring you to an empty view:
image

This view provides you with all the details and two edit cards:

image

image

Where saving provides you with a confirmation dialog:

image

Editing amounts is a bit bulkier:
image

image

But gives a nice "save overview" to help it:

image

UPDATE:

also amounts:

image

image

image

#121 

## Types of changes


apps/dashboard/src/modules/financial/composables/useTransactionCard.ts
```diff
@@ -0,0 +1,13 @@
+import { computed, ref } from 'vue';
+import type { TransactionResponse } from '@sudosos/sudosos-client';
+import { useTransactionStore } from '@/stores/transaction.store';
+import { isAllowed } from '@/utils/permissionUtils';
+
+export function useTransactionCard(transactionId: number) {
+  const transactionStore = useTransactionStore();
+  const transaction = computed(() => transactionStore.getTransaction(transactionId) as TransactionResponse);
+  const canEdit = computed(() => isAllowed('get', ['all'], 'Transaction', ['any']));
```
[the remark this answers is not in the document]
SuperVK: Had the same question

SuperVK (commented):
Looks good! A few comments from copilot are fair, and also the question why adding new products is not possible yet?

Also, in my head transactions editing is under financial, instead of admin. In my head admin is for everything in SudoSOS that is not related to finances (I think the maintainer overview should also be under admin). Admin should also mainly be for board (and team SudoSOS perhaps).

RubenLWF:
Just a small concern, do we want to display to users in some way that a transaction has been edited? Like a small annotation in the transaction overview?

JustSamuel:
If we return the `version` we could add a exclamation indicator with "this transaction has been modified."

RubenLWF:
Sounds like a good solution, but we might as well make a separate issue for that

SuperVK:
image
I cannot seem to select a new product?

apps/dashboard/src/modules/admin/navbar.ts
```diff
@@ -21,6 +24,11 @@ export function useAdminNav() {
             route: '/banner',
             visible: isAllowed('get', ['own'], 'Banner', ['any']),
           },
+          {
+            label: t('common.navigation.transactions'),
+            route: '/admin/transactions',
```
SuperVK:
```suggestion
            route: '/financial/transactions',
```

apps/dashboard/src/modules/financial/components/transaction/forms/TransactionAmountsBaseForm.vue
```diff
@@ -0,0 +1,277 @@
...
+  } else {
+    currentAmounts.push({ subTransactionIndex, rowIndex, amount });
+  }
+
+  props.form.context.setFieldValue('updatedAmounts', currentAmounts);
+}
+
+function handleProductSelection() {
+  // This will be called when the select value changes
+}
```
SuperVK: Why is this function empty?

SuperVK (approved):
I do not have my laptop to test, but the code looks good, so I say we can test in beta!

PR #650, opened 2025-10-29
Refetch buyer when setting in cart store

JustSamuel:
# Description

Some bugs turn out to be caused by old or partial user data (i.e. no GEWIS id). This also happens if we force a BaseUserResponse as a buyer. Refetching the user seems to be the easiest solution

## Types of changes


SuperVK (approved):
Makes sense, hopefully these issues will be fixed by the removal of GewisUserResponse and switching to external user id. Also food for thought, we now only need to pass the id instead of the whole object. 

Also see the low confidence comment in the response of copilot, it is a good point imo

PR #660, opened 2025-11-15
Implement QR based POS authentication

### LINKED ISSUE #369: [Feature] POS users

Point of sale

### What would you like?

We should implement the POS user authentication from https://github.com/GEWIS/sudosos-backend/issues/62. As also described in https://github.com/GEWIS/sudosos-backend/pull/232.

This means the flow should be as follows:
After loading the pos website, you are prompted with the same login as the dashboard. After logging in with your user account, you can log into one of the POS'es you own. Receiving a token for the POS that you just logged into. After logging into a POS, the normal keypad appears. Now product queries, and user queries should be made with the POS token, and the transaction should be made with the user token after the using the pin code.

### Why is this needed?

Because then every user cannot get every other user anymore, and admins can no longer make admin request with their pin code, making the system more secure. 

### Other information

We should probably take the login screen logic away from the dashboard, and put this in some sort of common place.

JustSamuel:
# Description
Implements the complete QR based pos authentication where.

#369 

## Types of changes


SuperVK (approved):
Looks good to me!
Points from the Signal:
- Allow opening POSes via an URL, to allow for default POS for after start up of the POS screens 
- Allow POS without pin auth to be logged in directly without first doing a single pin login.

PR #665, opened 2025-11-16
Derive placement of BuyerSelectButton from index

JustSamuel:
# Description

Instead of FIFO'ing the buttons, we now us their index to assign consistent spacing.

image

And now if we delete someone in the new UI it wil look like:

image

whilst in the old ui it would simply be:

image

## Types of changes


RubenLWF (approved):
LGTM! 🚀 

I assume you have also tested that new associates fill up the existing gaps properly?

PR #666, opened 2025-11-17
Add basic seller table in financial module

JustSamuel:
# Description

Start of trying to improve insight into sellers by adding a central point to navigate to them from a management perspective:

image

image

## Types of changes

### DIFF
--- a/apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
+++ b/apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
@@ -1,0 +1,119 @@
+<template>
+  <CardComponent class="w-full" :header="t('modules.financial.seller.title')">
+    <DataTable
+      data-key="id"
+      lazy
+      :loading="isLoading"
+      :paginator="true"
+      :rows="rows"
+      :rows-per-page-options="[5, 10, 25, 50, 100]"
+      table-style="min-width: 50rem"
+      :total-records="totalRecords"
+      :value="rowValues"
+      @page="onPage"
+    >
+      <Column field="id" :header="t('common.id')">
+        <template #body="slotProps">
+          <Skeleton v-if="isLoading" class="h-1rem my-1 surface-300 w-6" />
+          <span v-else>{{ slotProps.data.id }}</span>
+        </template>
+      </Column>
+      <Column field="name" :header="t('common.name')">
+        <template #body="slotProps">
+          <Skeleton v-if="isLoading" class="h-1rem my-1 surface-300 w-6" />
+          <UserLink v-else new-tab :user="slotProps.data.user" />
+        </template>
+      </Column>
+      <Column field="active" :show-filter-match-modes="false">
+        <template #header>
+          <div class="items-center flex flex-row gap-2">
+            {{ t('common.active') }}
+            <Checkbox v-model="showAllActiveSellers" v-tooltip="tooltipText" binary />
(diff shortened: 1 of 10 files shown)


apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
```diff
@@ -0,0 +1,119 @@
...
+        <template #body="slotProps">
+          <Skeleton v-if="isLoading" class="h-1rem my-1 surface-300 w-6" />
+          <UserLink v-else new-tab :user="slotProps.data.user" />
+        </template>
+      </Column>
+      <Column field="active" :show-filter-match-modes="false">
+        <template #header>
+          <div class="items-center flex flex-row gap-2">
+            {{ t('common.active') }}
+            <Checkbox v-model="showAllSellers" v-tooltip="tooltipText" binary />
```
RubenLWF: Can this be renamed to `showAllActiveSellers`? To make it a bit more clear what is omitted if this is not checked

apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+  <CardComponent class="w-full" :header="t('modules.financial.seller.title')">
+    <DataTable
+      class="w-full"
```
RubenLWF: This `class` attribute is not needed

apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
```diff
@@ -0,0 +1,119 @@
...
+      :rows="rows"
+      :rows-per-page-options="[5, 10, 25, 50, 100]"
+      table-style="min-width: 50rem"
+      :total-records="totalRecords"
+      :value="rowValues"
+      @page="onPage"
+    >
+      <Column field="id" :header="t('common.id')">
+        <template #body="slotProps">
+          <Skeleton v-if="isLoading" class="h-1rem my-1 surface-300 w-6" />
```
RubenLWF, line 17: Adding these skeletons is not needed since the table itself has a `:loading` attribute.

apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
```diff
@@ -0,0 +1,119 @@
...
+  isLoading.value = true;
+  try {
+    const response = await apiService.user.getAllUsers(
+      rows.value,
+      skip.value,
+      '',
+      showAllActiveSellers.value ? undefined : true,
+      undefined,
+      undefined,
+      'ORGAN',
```
RubenLWF: Change this to `USER_TYPES.ORGAN`

RubenLWF (changes requested):
A few small comments.

apps/dashboard/src/modules/financial/components/seller/SellerTable.vue
JustSamuel: Not really:

- the skeletons are still visable.
- datatable will try to render the columns otherwise, even with "is-loading" true, so you need a fallback when loading to prevent undefined errors.

RubenLWF: Makes sense

RubenLWF (approved):
🚀

PR #670, opened 2025-11-17
Add pagination to BannersList.vue

### LINKED ISSUE #51: Handle Pagination  on BannersView
Right now pagination is not being taken care of properly in the ProductsContainersView, instead just yeeting a big API call and calling it a day. This should be done more neatly.

JustSamuel:
# Description

Title.

Closes #51 

## Types of changes

### DIFF
--- a/apps/dashboard/src/composables/paginatedBanners.ts
+++ b/apps/dashboard/src/composables/paginatedBanners.ts
@@ -1,0 +1,71 @@
+import { ref, watch, onMounted } from 'vue';
+import type { BannerResponse } from '@sudosos/sudosos-client';
+import type { DataTablePageEvent } from 'primevue/datatable';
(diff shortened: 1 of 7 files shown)


JustSamuel:
First merge #665 to fixs typing of new client

SuperVK (commented):
It would be nice if the banners can refetch after it is updated in a dialog, like the current version does do already

SuperVK (approved):
I was thinking; with your approach, we can take out the entire store, as we are not doing any caching eitherway. And do all the API calls directly in the composable as abstraction layer for the views

JustSamuel:
> I was thinking; with your approach, we can take out the entire store, as we are not doing any caching eitherway. And do all the API calls directly in the composable as abstraction layer for the views

In this case we could, but I think its fine to leave it as is tbf.

PR #679, opened 2025-11-22
feat: url based login

SuperVK:
# Description
The ability to login with ?apiKey=XXX&userId=XXX&posId=XXX for the start up script of the POS screens.

## Types of changes


SuperVK:
As discussed in signal, we will get the pos jwt directly with the API key of the pos user. Depends now on #678

JustSamuel (approved):

PR #699, opened 2025-12-17
feat: Wrapped focus overlay

RubenLWF:
# Description

Adds an overlay to the first Wrapped card that only appears when the window is not in focus.

As soon as the window is focussed, the overlay disappears and Wrapped is shown as normal.

Nothing changes for desktop view.

Has been tested on mobile on Android in Firefox and Chrome.

## Types of changes


JustSamuel (approved):
✨ vibes ✨

PR #703, opened 2025-12-20
feat: inactive administrative cost overview

### LINKED ISSUE #651: [Feature] Administrative costs for inactive users
Implement a view with a wizard to handout and handle administrative costs.

JustSamuel:
# Description

## Types of changes


lib/nginx/default.conf
```diff
@@ -128,5 +128,8 @@ server {
     add_header 'Access-Control-Allow-Origin' '*';
     add_header 'Access-Control-Expose-Headers' '*';
     proxy_pass http://backend/;
+    proxy_buffer_size 128k;
+    proxy_buffers 4 256k;
+    proxy_busy_buffers_size 256k;
```
SuperVK, line 133: why this?

SuperVK (commented):
- I think the handout inactive administration cost from greyed out to solid makes more sense than appearing.
- The delete modal pops up twice, and deleting it twice creates 2 credits, essentially giving the person the original value extra.
- Deleting from the modal has a double click confirm, but then also shows 2 modals stacked

lib/nginx/default.conf
JustSamuel: Because of response sizes getting to large.

SuperVK (approved):

PR #713, opened 2026-01-07
fix: icon for logout

tomudding:
# Description
When a POS is not in `borrelMode`, a logout button is shown. However, the icon may suggest it is for clearing the selected products rather than logging out.

With this change, the icon is updated to signal that the action will log the user out.

## Types of changes


tomudding:
Bruh, GitHub added the extra newline through the online editor, not me.

SuperVK (approved):

PR #721, opened 2026-01-11
fix: disable authenticated POS login

RubenLWF:
# Description

Removes the need for users to log in using their pin after switching to a POS that does not require authentication.

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/components/SettingsIconComponent.vue
+++ b/apps/point-of-sale/src/components/SettingsIconComponent.vue
@@ -58,7 +58,9 @@ const handleSwitchToPos = async (pos: PointOfSaleResponse) => {
   if (pointOfSale.value?.id === pos.id) return;
   visible.value = false;
   await switchToPos(pos);
-  await logoutService();
+  if (pos.useAuthentication) {
+    await logoutService();
+  }
 };
 
 const forceExit = async () => {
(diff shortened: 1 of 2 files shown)


JustSamuel:
This doesn't really remove the need to login. This would only "reuse" the non-lesser token that the user got from the QR-code. However keeping this "running" is also not preferred.

The main idea I had was to enable POS tokens of borrelmode pos to be able to create transactions and the rest. This would allow the POS token itself to make transactions, but this might not be super easy to implement in this ducttape situion.

RubenLWF:
> The main idea I had was to enable POS tokens of borrelmode pos to be able to create transactions and the rest. This would allow the POS token itself to make transactions, but this might not be super easy to implement in this ducttape situion.

We should also keep in mind that it should be possible to switch to another POS from a borrelmode POS. This might be a bit complicated if we do not have user JWT token.

JustSamuel:
Then I saw we close this for now until we have a better plan on how to deal with the stated issues?

SuperVK:
> We should also keep in mind that it should be possible to switch to another POS from a borrelmode POS. This might be a bit complicated if we do not have user JWT token.

You can just let that fly, it is not something that is used (I think at least). Or you can make it such that lesser tokens can request new POS tokens?

JustSamuel:
> You can just let that fly, it is not something that is used (I think at least). Or you can make it such that lesser tokens can request new POS tokens?

This is used every borrel? Or do we want the flow to be "force logout + scan the qr code again"? Because I think people (BAC) will consider this a degredation.

RubenLWF:
I agree with closing this for now, I have create https://github.com/GEWIS/sudosos-frontend/issues/727 to further discuss this.

PR #724, opened 2026-01-12
Move pos overlay to view

### LINKED ISSUE #709: [Refactor]: No Pos Overlay to auth guard
Currently, NoPosOverlay.vue is overlayed on top of all the current content. This can also be done with an auth guard, this is more robust way to do this as there are no hidden assumptions, similar to the dashboard. This makes sure we do not stack multiple views on top of each other.

JustSamuel:
# Description

Simple refactor that moves the overlay logic to route guards.

Closes #709


SuperVK (commented):
I would delete @/components/NoPosTokenOverlay.vue in this PR as it is no longer used

SuperVK (approved):

PR #728, opened 2026-01-16
feat(dashboard): add notification preferences card

JustSamuel:
# Description

Notification preferences
image

## Types of changes


SuperVK (commented):
Wouldn't it be more logical to match it closer with the backend. Like two switch buttons with 

**Notifications for your account:**
"Transactions made by myself [switch]" 
"Transactions made by others [switch]"


Or maybe even simpler just "No notifs, all notifs, only by others", as that is the most likely?

I think like this is also fine, but think the above is a bit more clean

SuperVK (approved):
I would change "Notification preferences" to "Email notification preferences", but then it is good!

PR #745, opened 2026-01-21
fix: prevent nfc pin login

RubenLWF:
# Description

The nfc code was put in as a pin code because of the keypress listeners.

## Types of changes


JustSamuel (approved):
Surely.

If the shareholders approve, I approve 🫡

PR #768, opened 2026-01-25
feat(pos): add click sounds

RubenLWF:
# Description

Plays a sound every time a clickable element is pressed. Clickable elements are:
 - Buttons
 - Elements with an HTML `onClick` property
 - Elements with a Vue `onClick` property

## Types of changes


JustSamuel (approved):
🦫

PR #774, opened 2026-01-29
fix(dashboard): beta sync overlay spinner color

RubenLWF:
# Description

The beta sync overaly background color is the primary color, which made it so that the spinner was not visible, since it is the same color.

The spinner is now white.

Before:
image

After:
image

## Types of changes


CodeNamedRobin (approved):
Looks good to me!

PR #789, opened 2026-02-22
refactor: linking to entities renewed

SuperVK:
# Description

I'm changing the way we link to other pages once more, for two reasons, inspired by parelpracht, and the rest of the internet.
1. Forcing the new tab is not ideal, browsers have a way to force a new tab, but not the other way around. So if you do want to open in the same tab that is not possible. This PRs switches back to forcing in the same tab, with the possibility to open in a new tab by using right click, middle click or ctrl click. 
1. Instead of opening new places via an action button, it works by linking the actual entities. This is a more intuitive way to do this, as you are actually clicking the name of where you are going. Especially the POS overview page explains why this works well, as instead of only being able to open the POS, you can also now open the owner:
image
vs
image

Other examples:
Screenshot From 2026-02-22 17-51-47
image

As a nice QoL it also allows directly linking to users on the GEWIS page. 

## Types of changes

- UI/UX

### DIFF
--- a/apps/dashboard/src/modules/admin/components/users/AdminUserTable.vue
+++ b/apps/dashboard/src/modules/admin/components/users/AdminUserTable.vue
@@ -36,10 +36,10 @@
       <template v-else #body="slotProps">
         <a
           v-if="slotProps.data.gewisId"
-          :href="`https://gewis.nl/member/${slotProps.data.gewisId}`"
-          target="_blank"
-          rel="noopener noreferrer"
           class="flex items-center gap-2 cursor-pointer hover:opacity-80 text-primary"
+          :href="`https://gewis.nl/member/${slotProps.data.gewisId}`"
+          rel="noopener noreferrer"
+          target="_blank"
         >
           {{ slotProps.data.gewisId }}
           <i class="pi pi-external-link text-sm" />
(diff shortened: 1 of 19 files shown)


apps/dashboard/src/modules/admin/components/users/AdminUserTable.vue
```diff
@@ -33,16 +33,26 @@
...
+        <a
+          v-if="slotProps.data.gewisId"
+          class="flex items-center gap-2 cursor-pointer hover:opacity-80 text-primary"
+          :href="`https://gewis.nl/member/${slotProps.data.gewisId}`"
+          rel="noopener noreferrer"
+          target="_blank"
+        >
+          {{ slotProps.data.gewisId }}
+          <i class="pi pi-external-link text-sm" />
+        </a>
```
RubenLWF: suggestion:

Add an `ExternalLink` component, since this GEWIS ID link is used in multiple places.

RubenLWF (changes requested):

apps/dashboard/src/modules/admin/components/users/AdminUserTable.vue
SuperVK: Done

JustSamuel (approved):

RubenLWF (approved):

PR #790, opened 2026-02-22
fix: improve user table

SuperVK:
# Description

A few quality of life fixes for the user table:
1. Improved searching. Diacritics are now ignored; you can search for Noel and find Noël. Especially useful for non-Dutch/English European last names. Furthermore, results are not filtered by Fuse anymore, only sorted. Allows for more powerful searches while still keeping the relevance sorting from Fuse. It is also possible now to search by nickname and email. 
1. Improved filtering. The default is more sane now, namely Active and both 18+ and 18- by default. You can also filter on inactive, or no filtering. Same for 18+/18-

## Types of changes

- UI/UX


CodeNamedRobin (approved):
LGTM!

RubenLWF (approved):

PR #804, opened 2026-03-20
Delete all Grimbergen products

### LINKED ISSUE #783: [Discussion]: Add April Fools steekproef checkout
### What would you like?

Add an April Fools “steekproef” prank during checkout in the SudoSOS frontend, inspired by Albert Heijn: with a small random chance, the checkout flow is blocked and the user is told to get a board member to verify the purchase before continuing.

This should be a fun, non-destructive easter egg and should be easy to enable/disable (and ideally only active around April 1st).

--- 

What are we going to do for April 1st?

Do we want to do something funny on the POS? Like sometimes showing a "Een medewerker komt zo bij u" like at AH selfscan?

Image

RubenLWF:
# Description

Adds the steekproef modal, which has a 50% chance of appearing when checking out on April Fool's day.

A check to see whether it is currently April Fool's day has been added to the settings store, to make sure it only appears on the desired date.

The modal can be closed (which will continue the checkout process) by solving a CAPTCHA (a simple math question) that appears after 5 seconds. If the user fails the CAPTCHA too often, 3 times, the user will be logged out.

Also adds an age verification modal that appears when a user attempts to buy an alcoholic product. If the user presses "No", all alcoholic items are removed from the cart.

### Screenshot:
image

image

https://github.com/GEWIS/sudosos-frontend/issues/783

## Types of changes


apps/point-of-sale/src/components/AprilFoolsComponent.vue
JustSamuel, line 1: Maybe add a sound?

JustSamuel (approved):
We should also make sure the buyer is of age

JustSamuel (approved):

PR #805, opened 2026-03-22
Split organ transactions into tabs

JustSamuel:
# Description
Transactions of Organs provide less info than the tranfers. So we un-unify this table and add tabs instead. Also fix the rendering issue of payouts.

image
image
image

## Types of changes


SuperVK (approved):
Agree with AI

PR #820, opened 2026-04-07
chore: lint

CodeNamedRobin:
Fixes several lint issues

Adds claude to gitignore

nog vragen

## Types of changes

### DIFF
--- a/apps/dashboard/src/components/FindRole.vue
+++ b/apps/dashboard/src/components/FindRole.vue
@@ -27,10 +27,11 @@ const emits = defineEmits(['update:role']);
 const props = defineProps({
   role: {
     type: Object as PropType<RoleResponse>,
+    default: undefined,
   },
   default: {
     type: Object as PropType<RoleResponse>,
-    required: false,
+    default: undefined,
   },
   placeholder: {
     type: String,
--- a/apps/dashboard/src/modules/admin/components/rbac/RBACTable.vue
+++ b/apps/dashboard/src/modules/admin/components/rbac/RBACTable.vue
@@ -255,6 +255,7 @@ const { t } = useI18n();
 const props = defineProps({
   state: {
     type: Object as PropType<string>,
+    default: undefined,
   },
   form: {
     type: Object as PropType<Form<yup.InferType<typeof rbacSchema>>>,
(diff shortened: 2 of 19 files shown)


apps/dashboard/src/components/FindRole.vue
```diff
@@ -27,10 +27,12 @@ const emits = defineEmits(['update:role']);
 const props = defineProps({
   role: {
     type: Object as PropType<RoleResponse>,
+    default: undefined,
```
RubenLWF, line 30: question:

Do we need this?

I think this is already done implicitly

apps/dashboard/src/components/FindRole.vue
```diff
@@ -27,10 +27,12 @@ const emits = defineEmits(['update:role']);
...
   },
   default: {
     type: Object as PropType<RoleResponse>,
     required: false,
+    default: undefined,
```
RubenLWF, line 34: question:

Do we need this?

I think this is already done implicitly

apps/dashboard/src/modules/admin/components/rbac/RBACTable.vue
```diff
@@ -255,6 +255,7 @@ const { t } = useI18n();
 const props = defineProps({
   state: {
     type: Object as PropType<string>,
+    default: undefined,
```
RubenLWF, line 258: question:

Do we need this?

I think this is already done implicitly

RubenLWF (changes requested):
Few small questions

apps/dashboard/src/components/FindRole.vue
CodeNamedRobin: I disabled the rule for this line now, since it seems to be broken

apps/dashboard/src/components/FindRole.vue
CodeNamedRobin: I disabled the rule for this line now, since it seems to be broken

apps/dashboard/src/modules/admin/components/rbac/RBACTable.vue
CodeNamedRobin: I disabled the rule for this line now, since it seems to be broken

RubenLWF (approved):

PR #826, opened 2026-04-20
Incorrect VAT calculation in useInvoiceEntries

CodeNamedRobin:
Calculates VAT by individual product, not invoice entry

## Types of changes


RubenLWF (approved):
good catch!

PR #828, opened 2026-05-07
Show deactivated account modal on POS login

### LINKED ISSUE #813: 🎨 Design: Show message for deactivated users
### What would you like?

Improve Point-of-Sale frontend UX for users whose account is deactivated (deleted=false, active=false).

Context:
- Reported scenario: user can authenticate/log in, but cannot buy stuff or perform actions.
- Expected: the POS should clearly inform the user their account is deactivated instead of leaving them in a confusing partially-logged-in state.

Relevant area explored:
- POS app lives in `apps/point-of-sale/`.
- POS API service is created in `apps/point-of-sale/src/services/ApiService.ts` using `/api/v1`.

Notes:
- This issue is about frontend behavior/messaging; exact backend error/response that indicates deactivation still needs to be confirmed in the POS flow.

### Why is this needed?

Deactivated users currently experience confusing behavior (can log in but cannot buy). This leads to poor UX and increases support burden; the UI should clearly communicate account status.

### How could it be implemented?

Add a clear deactivated-account handling path in the POS frontend: when the backend indicates the user is inactive/deactivated, show a dedicated error state/message and block further actions. Where possible, detect this as early as login/session initialization and avoid entering the normal POS flow.

Implementation details intentionally left open until the exact API response/error is identified.

### Other information

Duplicate search performed in GEWIS/sudosos-frontend: no existing issues found with similar keywords.

JustSamuel:
# Description

A user with `active=false` can authenticate on the POS but can't buy anything once they're in. They get parked on the cashier screen with nothing telling them why nothing works.

This PR pops a modal right after auth if `user.active === false`. The modal says the account is deactivated and to contact the SudoSOS administrators. "I Understand" logs them out and sends them back to the keypad, so they don't enter the cashier flow at all.

## What changed

- New `DeactivatedAccountModalComponent.vue`, modeled on `TopUpWarningComponent` (closable=false, dismissable-mask=false, close-on-escape=false, same "I Understand" button styling). One-way `:show` + `@dismiss`; you can only get out by clicking the button.
- `useLoginForm.ts` checks `user.active` after a successful login. If false, it sets `showDeactivatedModal=true` and skips the `/cashier` navigation. Adds `dismissDeactivatedModal()` which logs out and resets the keypad.
- `LoginView.vue` mounts the modal and wires the dismiss handler.

The login call still completes (so a JWT briefly exists), but the logout fires on dismiss before any navigation off `/login`.

## Screenshot

Account deactivated modal on the POS login screen

Closes #813

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/components/DeactivatedAccountModalComponent.vue
+++ b/apps/point-of-sale/src/components/DeactivatedAccountModalComponent.vue
@@ -1,0 +1,52 @@
+<template>
+  <Dialog
+    class="w-[35rem]"
+    :closable="false"
+    :close-on-escape="false"
+    :dismissable-mask="false"
+    :draggable="false"
+    header="Account deactivated"
+    modal
+    :visible="show"
+  >
+    <Message :closable="false" :icon="undefined" severity="warn">
+      Your account is <span class="font-bold">deactivated</span>.
+      <br />
+      You cannot make purchases at the POS until it is reactivated.
+      <br /><br />
+      Please contact the SudoSOS administrators to resolve this.
+    </Message>
+
+    <div class="flex justify-center items-center mt-4">
+      <Button class="text-xl px-6 py-3 understand-button" @click="emit('dismiss')"> I Understand </Button>
+    </div>
+  </Dialog>
+</template>
+
+<script setup lang="ts">
+defineProps({
+  show: {
+    type: Boolean,
+    required: true,
+  },
+});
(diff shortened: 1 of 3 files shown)


RubenLWF (approved):

tomudding:
There is no guidance on who the "SudoSOS administrators" are or how to contact them; I think that would have been better for the UX.

Issue #817, opened 2026-04-03
[Feature]: Unified entity view system with query-param filter persistence


JustSamuel:
### What would you like?

Establish a standardised "entity view" pattern across the dashboard so that every major entity (Transaction, Transfer, Invoice, Payout Request, Seller Payout, Fine, Write-off, etc.) has a consistent representation: same column conventions, same filter controls, and filter state persisted in and driven by URL query parameters.

### Why is this needed?

Currently each entity table is implemented independently with different layouts, different filter UX, and no URL state — refreshing the page loses your filters, and there's no way to link a colleague directly to a filtered view. As more transfer-type entities are added, inconsistency compounds. A unified system:

- Makes every table instantly familiar to users
- Makes filtered views shareable via URL
- Makes it dramatically easier to add new entity views in the future
- Enables navigation between related views with filters pre-applied (e.g. "show me all transfers for this user" from the user detail page)

### How could it be implemented?

This builds on and extends the composable unification work in #807 (which already targets invoice/payout/seller-payout DataTables). The scope here is broader: a full design system for entity views.

#### Core composable: `useEntityTable`

Extend `useDataTableYear` (from #807) into a general `useEntityTable` composable that manages:

- **Fiscal year selection** — tab-based, synced to `?year=` query param
- **Search** — by ID or name, synced to `?search=` query param  
- **Status/type filters** — synced to `?filter[status]=` query params
- **Pagination** — page + rows-per-page, synced to `?page=` and `?rows=`
- **Sort** — column + direction, synced to `?sort=` and `?dir=`

All query params are read on mount (so a shared URL restores full state) and written on change (so the URL always reflects current view).

#### Standard entity view structure

Each entity view should follow the same visual structure:
1. **Header** — entity name, total count, optional export button
2. **Filter bar** — year tabs, search input, status/type dropdowns (only what's relevant per entity)
3. **DataTable** — consistent column ordering: ID → date → counterparty → amount → status → actions
4. **Detail navigation** — clicking a row opens the detail view (or a side panel), preserving table state in the URL so the back button works

#### Cross-page filter passing

When navigating from one view to another with a pre-applied context (e.g. from a user's profile to "all their transactions"), the target URL should include the relevant filter params so the table opens pre-filtered. This replaces ad-hoc store state passing.

#### Entities to cover

In priority order (most-used first):
- [ ] Transaction
- [ ] Transfer
- [x] Invoice
- [x] Payout Request
- [ ] Seller Payout
[...]

Issue #506, opened 2025-02-19
[Discussion]: What is needed for from the frontend for a good financial audit?


JustSamuel:
During my 26 hours of fixing the bookkeeping, I had to use a lot of db queries, which was fine since I can access the db, but for years after me this information should be provided by the dashboard. Therefore, I am suggesting creating a "financial insights" tab / module that helps doing such an audit without db queries. 

The main goal in this tab is to be able to quickly check numbers from our bookkeeping with those in SudoSOS. The scope is therefore mostly to present numbers for a bookyear / quartile. 

The main numbers are
- The top-ups
- The total writeoffs
- The total payouts
- The total seller payouts
- The total amount of invoices
- The total amount of fines

These numbers together describe how the balance in SudoSOS changed. 

To be able to also nicely check these numbers, we should be able to export these list to an csv, but that will be a backend issue.

Jobbiesaus:
I completely agree with Samuel on this. He had to take time out of his "afstuderen" to help me get the correct numbers, I do not think the ABC should be involved in finances.

Furthermore, all of these should be possible to accessed with exact dates.

tomudding:
As mentioned in the boardroom, I think this is an excellent idea and a wise thing to implement.
  
> * The total amount of fines

Make sure that the amount of waived fines is also very clear to avoid any confusion in that regard.

I do not know whether it should be part of this overview or something else, but today, we noticed an issue in the fine -> transfer mapping because of a lack of foreign key constraints. That is also something you ideally want to be able to see (however, if we can ensure/be sure that this never happens again, we do not need such functionality).

Another thing: How would we represent refunds? We have not had to deal with that yet, but it is something to consider.

CodeNamedRobin:
@RubenLWF and I have thought about this, and think the best option for this is to create a `DataTable` with every single row that affects the SudoSOS balance. This table should have a customizable date range, and should show the total balance within the date range. It should also be possible to filter the type of rows, such as Invoice Transfers, Monthly Sales, Topups, Write-offs, etc. This is all designed to closely match MultiVers, such that the KCC can easily check this against the BAC Bookkeeping. It might also be nice to have links to the full entity, depending on the type of row. This could be that an invoice row links to the particular invoice. 
I will take on this issue together with @RubenLWF, pair programming our way to victory, and by victory I mean an easy Financial Audit on the BAC Bookkeeping.

SuperVK:
From the kasco meeting rn:
[...]

PR #829, opened 2026-05-12
feat: transfer view

CodeNamedRobin:
Provides a transfer view for each category of transfer. This enables the BAC PM and Financial Audit Committee to easily check the SudoSOS balance, if it tracks correctly with the MV balance and if it doesn't, where the fault lies. 
Each view provides a list of individual transfers of that category with a date, credit, debit, saldo, id and description. Each view also provides a widget showing the aggregate of all the transfers in that category. The user can select transfers to exclude from the total, allowing fine-grained checking with MV. 
image

Works on #817
Works on (and potentially closes) #506 
## Types of changes


RubenLWF (changes requested):
sjaars

RubenLWF:
Please have a look at the copilot comments

JustSamuel:
Maybe also add which type of transfer it is?

CodeNamedRobin:
> Maybe also add which type of transfer it is?

I did think of this, but might be irrelevant due to it being in the route? What I implemented are `/financial/transfer/{category}`, so the type will be all the same. I could add a `/financial/transfer/all` route that does specify type for each individual transfer.

JustSamuel:
Maybe then improve  visually what type it is? since now screenshots are ambiguous.

PR #834, opened 2026-06-05
fix: only show admin dashboard with correct permissions

RubenLWF:
# Description
Users who did not have permission to make all the calls required to populate the admin dashboard were redirected anyway. This made it so that they were shown an (almost) empty dashboard, which is not useful.

## Types of changes


CodeNamedRobin (approved):
LGTM!

PR #836, opened 2026-06-05
feat: use TOS from backend endpoints

RubenLWF:
# Description

Moves to using the new `GET /terms-of-service/latest` endpoint to fetch the latest version of the ToS, rather than using a `.md` file stored in the FE.

image

image

image

## Types of changes


apps/dashboard/src/modules/auth/views/AuthTermsOfServiceView.vue
```diff
@@ -31,18 +31,19 @@ import { useAuthStore, useUserStore } from '@sudosos/sudosos-frontend-common';
...
+import { useTermsOfServiceStore } from '@/stores/termsOfService.store';
 
 const { t } = useI18n();
 
 const authStore = useAuthStore();
 const userStore = useUserStore();
+const tosStore = useTermsOfServiceStore();
 
-const tos = marked(termsOfService);
+const tos = marked(tosStore.getTermsOfService);
```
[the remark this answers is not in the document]
RubenLWF: I don't think we need this.

apps/dashboard/src/components/footer/FooterTermsOfServiceModal.vue
```diff
@@ -22,10 +19,12 @@ import Dialog from 'primevue/dialog';
...
 import { marked } from 'marked';
 import { useI18n } from 'vue-i18n';
-import termsOfService from '@/locales/termsOfService.md?raw';
+import { useTermsOfServiceStore } from '@/stores/termsOfService.store';
+
+const tosStore = useTermsOfServiceStore();
 
 const { t } = useI18n();
-const tos = marked(termsOfService);
+const tos = marked(tosStore.getTermsOfService);
```
JustSamuel: This means it's not reactive. Could not this not lead to an empty ToS if internet is slow?

apps/dashboard/src/utils/beforeLoadUtil.ts
```diff
@@ -26,4 +28,8 @@ export default async function beforeLoad() {
     const authStore = useAuthStore();
     authStore.logout();
   });
+
+  await termsOfServiceStore.fetchTermsOfService().catch((e) => {
+    console.error(e);
```
JustSamuel, line 33: should wen to have an error state? or at least disable "Accept tos" if you were never able to read it?

JustSamuel (approved):
Small points

PR #850, opened 2026-07-30
Pay for an anonymous cart with a Stripe terminal

JustSamuel:
# Description

Lets someone without a SudoSOS account pay for a cart straight on a Stripe
Terminal reader. This is the frontend half of GEWIS/sudosos-backend#967, which
added anonymous terminal payments in time for the Introduction Week.

In borrel mode a card button shows up next to the checkout button. Pressing it
registers the cart as a `TerminalPayment`, hands it to the reader, and polls
until the payment settles. On success the backend books the transaction and the
cart clears the same way it does after any other sale.

Only borrel mode gets the button. An anonymous sale needs a cart with nobody
selected, and "select no one" only exists on an unauthenticated POS. On an
authenticated POS there is always someone to charge, so the button would sit
there greyed out forever.

A few decisions that need explaining.

`from` and `createdBy` are both the POS user. That matches how an anonymous sale
is supposed to work, and it keeps the payment inside the POS token's `own` RBAC
relation for its whole life. Cancelling detaches the temporary transaction, so
with the cashier as `createdBy` there would be nothing left for `getRelation` to
match on, and polling would start coming back 403.

The reader is resolved before the payment gets created. Creating a
`TerminalPayment` also creates a temporary transaction and a Stripe payment
intent, and there is no reason to do that when there is nowhere to send it. If
handing the payment to the reader fails after that, we cancel it rather than
leave it stuck in `created`.

Everything goes through `posApiService`. Only the Point of Sale role has
`TerminalPayment` permissions, so a cashier's own token gets rejected.

Polling chains `setTimeout` instead of using `setInterval`, so a slow response
can't overlap the next request. It tolerates a few network errors in a row
before giving up, and treats 403/404 as a cancellation.

Servers without Stripe configured don't mount `/v1/terminal-payments` at all, so
that 404 is reported as "card payments are not enabled on this SudoSOS server"
rather than blamed on a missing reader.

## Client bump

Pins `@gewis/sudosos-client` to `0.0.0-develop.c63332b` (backend `c63332bf`),
the first build that exposes `TerminalPaymentsApi`. Nothing tagged has it yet,
so this wants moving to a real version once one is cut.

The bump drags two unrelated fixes along with it:

- axios goes to an exact `1.18.1` to match the client's own dependency. The old
  `^1.14.0` range resolved a second copy of axios, which matters because
  `lib/common` builds the axios instance and hands it to client classes typed
  against a different one. The repo already pins this way.
- `AcceptTosRequest` now requires a `version`, so `updateUserToSAccepted` takes
  it and the dashboard passes the version it rendered. The backend only accepts
  the current version, so sending what the user actually saw is better than
  re-fetching at submit time.

## Verification

Against a local backend and POS dev server: button visibility per mode, the
enablement matrix with its styling and tooltips, and the real 404 from a
backend without Stripe keys.

The happy path and the failure cases used a stub at the XHR boundary, so
everything above the network call is the real store, modal and component:

- `terminals -> create -> process -> poll -> paid`, checking the outgoing
  `TransactionRequest` field by field, the reader being picked automatically,
  then the cashout sound and a cleared cart
- zero and multiple available readers, which give different messages and create
  no payment at all
- cashier cancel: sends `DELETE`, keeps the cart, stops polling without leaking
  a timer
- a Stripe-side cancel picked up by polling, and a 403 mid-poll
- `process` failing, which cancels the payment it just created

Two gaps.

There are no unit tests, because this repo has no test runner at all and adding
the first one inside a payment feature felt like the wrong place for it. The
store is shaped to be testable whenever that happens.

Nothing has run against real Stripe yet. That needs test credentials plus either
hardware or Stripe's simulated reader. Worth doing before anyone trusts this at
a live bar; `cli/stripe-terminal.ts` in the backend is the place to start.

Frontend side of GEWIS/sudosos-backend#967, which implements scenario 1 of
GEWIS/sudosos-backend#852.

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/components/Cart/CartActionsComponent.vue
+++ b/apps/point-of-sale/src/components/Cart/CartActionsComponent.vue
@@ -28,7 +28,11 @@
     @denied="onAgeVerificationDenied"
     @update:show="showAgeVerification = $event"
   />
-  <div class="flex justify-between w-full">
+  <TerminalPaymentModal :show="showTerminalPayment" @update:show="showTerminalPayment = $event" />
+  <!-- gap-6 matches the spacing justify-between happens to leave between the
+       fixed-width Checkout button and the logout icon on an authenticated POS,
+       so borrel mode (where Checkout is full width) looks identical. -->
+  <div class="flex justify-between gap-6 w-full">
     <Button
       class="border-0 checkout font-medium rounder text-3xl"
       :class="{ countdown: checkingOut, disabled: !enabled, borrelMode }"
@@ -37,6 +41,15 @@
       {{ checkoutText }}
     </Button>
     <div class="flex justify-center items-center">
+      <Button
+        v-if="borrelMode"
+        aria-label="Pay with card terminal"
+        class="terminal p-3 text-2xl text-white flex items-center justify-center w-16 h-16"
+        :disabled="!terminalEnabled"
+        @click="payWithTerminal"
+      >
+        <i class="pi pi-credit-card" style="font-size: 2rem" />
+      </Button>
       <Button
         v-if="!borrelMode"
         class="p-3 text-2xl text-white flex items-center justify-center w-16 h-16"
@@ -57,6 +70,7 @@ import { useCheckoutTimer } from '@/composables/useCheckoutTimer';
 import { playAudio, Sound } from '@/utils/audioUtil';
 import AprilFoolsComponent from '@/components/AprilFoolsComponent.vue';
 import AgeVerificationComponent from '@/components/AgeVerificationComponent.vue';
+import TerminalPaymentModal from '@/components/Cart/TerminalPaymentModal.vue';
 
 const emit = defineEmits(['selectCreator']);
 
@@ -85,6 +99,21 @@ const checkoutText = computed(() => {
 const enabled = computed(() => {
   return cartItems.length > 0 && buyer.value;
 });
+
+const showTerminalPayment = ref(false);
+
+// Terminal payments are anonymous for now: the sale is booked in the name of
+// the POS itself. That only makes sense in borrel mode, where "select no one"
+// exists -- an authenticated POS always has someone to charge, so the button
+// is not rendered there at all (see the v-if in the template).
+const terminalEnabled = computed(() => cartItems.length > 0 && !buyer.value);
+
+const payWithTerminal = () => {
+  if (!terminalEnabled.value) return;
+
+  stopCheckout();
+  showTerminalPayment.value = true;
+};
 
 const logout = async () => {
   stopCheckout();
@@ -153,6 +182,13 @@ const checkout = () => {
 .clear {
   color: white;
   background-color: red;
+}
+
+.terminal:disabled {
+  background-color: grey;
+  border-color: grey;
+  opacity: 1;
+  cursor: not-allowed;
 }
 
 .checkout {
--- a/apps/point-of-sale/src/components/Cart/TerminalPaymentModal.vue
+++ b/apps/point-of-sale/src/components/Cart/TerminalPaymentModal.vue
@@ -1,0 +1,193 @@
+<template>
+  <Dialog
+    v-model:visible="visible"
+    class="w-120"
+    :closable="false"
+    :dismissable-mask="false"
+    :draggable="false"
+    header="Card payment"
+    modal
+    :pt="{
+      header: () => ({ class: ['dialog-header'] }),
+    }"
+  >
+    <div class="flex flex-col items-center gap-6 py-4">
+      <!-- `idle` is covered too: the dialog is visible for a frame before
+           startPayment flips the phase, and an empty dialog looks broken. -->
+      <template v-if="phase === 'idle' || phase === 'preparing'">
+        <ProgressSpinner />
+        <p class="text-xl text-center">Sending to the terminal...</p>
+      </template>
+
+      <template v-else-if="phase === 'waiting'">
+        <ProgressSpinner />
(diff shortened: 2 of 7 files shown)


docs/superpowers/specs/2026-07-30-pos-terminal-payment-design.md
RubenLWF, line 1: Suggestion:

Drop this commit

apps/point-of-sale/src/components/Cart/CartActionsComponent.vue
```diff
@@ -86,6 +101,27 @@ const enabled = computed(() => {
...
+// exists -- an authenticated POS always has someone to charge, so the button
+// is not rendered there at all (see the v-if in the template).
+const terminalEnabled = computed(() => cartItems.length > 0 && !buyer.value);
+
+const terminalTooltip = computed(() => {
+  if (cartItems.length === 0) return 'Add products to the cart first';
+  if (buyer.value) return 'Card payments are only available when no one is selected';
+  return 'Pay with card terminal';
+});
+
```
RubenLWF: Question:

Why the tooltips? These will never appear on the POS right?

apps/point-of-sale/src/components/Cart/TerminalPaymentModal.vue
```diff
@@ -0,0 +1,193 @@
...
+}>();
+
+const toast = useToast();
+const cartStore = useCartStore();
+const terminalPaymentStore = useTerminalPaymentStore();
+const { getPosUserIdFromToken } = usePosToken();
+
+const visible = computed({
+  get: () => props.show,
+  set: (value) => emit('update:show', value),
```
RubenLWF, line 80: Question:

What do we need these for?

apps/point-of-sale/src/stores/terminalPayment.store.ts
```diff
@@ -0,0 +1,241 @@
...
+  public constructor(
+    public readonly reason: 'none' | 'multiple',
+    public readonly available: number,
+  ) {
+    super(`Expected exactly one available Stripe terminal, found ${available}`);
+    this.name = 'TerminalUnavailableError';
+  }
+}
+
+const POLL_INTERVAL_MS = 2000;
```
RubenLWF, line 33: Question:

If we want to improve the speed of the payment appearing on the terminal, we can decrease this right? (We don't need to, since we have websockets in the pipeline anyway)

apps/point-of-sale/src/stores/terminalPayment.store.ts
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Resolve the single available Stripe reader for this POS.
+     * @throws {TerminalUnavailableError} when there is not exactly one.
+     */
+    async resolveTerminal(): Promise<StripePaymentTerminalResponse> {
+      const terminals = await posApiService.terminalPayments.getStripeTerminals().then((res) => res.data);
+      const available = terminals.filter((t) => t.available);
+
+      if (available.length !== 1) {
+        throw new TerminalUnavailableError(available.length === 0 ? 'none' : 'multiple', available.length);
```
RubenLWF: Suggestion:
Only pass `available.length`, decide whether to show `none` or `multiple` in the error itself.

RubenLWF (changes requested):
Big mf plans 🦫 

Not a big fan of the polling approach, but luckily a websocket solution is in the pipeline already 🥳 

Can you please have a look at the very specific AI generated comments?

docs/superpowers/specs/2026-07-30-pos-terminal-payment-design.md
JustSamuel: Addressed -- dropped both commits that added/edited this file. It was a planning scratch doc, not project documentation, and nothing else in the branch referenced it.

apps/point-of-sale/src/components/Cart/CartActionsComponent.vue
JustSamuel: Addressed -- you're right, this used the native `title` attribute, which only shows on mouse hover and would never appear on a touch POS. Removed it and the now-unused `terminalTooltip` computed; the disabled state already communicates that the button can't be pressed.

apps/point-of-sale/src/components/Cart/TerminalPaymentModal.vue
JustSamuel: `props`/`emit` here for `show`/`update:show` mirror the exact pattern already used by the sibling dialogs in this app (`AprilFoolsComponent`, `TopUpWarningComponent`, `AgeVerificationComponent`) -- all of them use `defineProps` + `defineEmits('update:show')` + a `computed({get, set})` proxy for `v-model:visible`, rather than `defineModel`. I kept it consistent with those rather than introducing a different pattern (the dashboard app does use `defineModel` for its dialogs, but that's not this app's convention). Happy to switch all four over to `defineModel` in a follow-up if you'd rather standardize on that going forward -- let me know.

apps/point-of-sale/src/stores/terminalPayment.store.ts
JustSamuel: Correct, and agreed -- leaving it at 2s since the websocket PR replaces this polling loop entirely rather than tuning it.

apps/point-of-sale/src/stores/terminalPayment.store.ts
JustSamuel: Addressed -- `TerminalUnavailableError` now takes just `available: number` and derives `reason` in its own constructor, so there's one place that decides none vs. multiple instead of two.

RubenLWF (approved):
Ja mooi

PR #851, opened 2026-07-31
Deduplicate dashboard mutation table components

JustSamuel:
# Description

The dashboard had four near-identical mutation table components (`MutationsBalance`, `MutationsPOS`, `MutationsSeller`, `MutationsTransfers`), each reimplementing the same table: lazy pagination, loading skeletons, the date column, amount color coding, an info-icon column, and the mutation/inactive-cost modal wiring. Small fixes had already drifted out of sync between variants - for example, only three of the four reset `isLoading` on a failed page fetch.

This PR pulls all of that shared behavior into one `MutationsTable.vue`. The four original components keep their existing public props (`getMutations`/`getTransfers`, `paginator`, `rowsAmount`, `preload`) and their own specific columns, but now just render `MutationsTable` and pass their extra columns through a scoped slot. I didn't have to touch any call sites.

Verified manually in the browser (admin, member, organ, and seller views) against a freshly seeded local backend: pagination, amount coloring (green for deposits, red for fines, default for transactions), the edited-icon on Balance, and the detail/inactive-cost modals all behave the same as before.

Closes #847

## Types of changes

### DIFF
--- a/apps/dashboard/src/components/mutations/MutationsBalance.vue
+++ b/apps/dashboard/src/components/mutations/MutationsBalance.vue
@@ -1,164 +1,52 @@
 <template>
-  <DataTable
-    lazy
+  <MutationsTable
+    ref="tableRef"
+    :fetch-page="fetchPage"
     :paginator="paginator"
-    :rows="rows"
-    :rows-per-page-options="[5, 10, 25, 50, 100]"
-    :total-records="totalRecords"
-    :value="mutations"
-    @page="onPage($event)"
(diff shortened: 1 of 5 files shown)


RubenLWF (approved):
I'm not reviewing this shit, does it work?

CodeNamedRobin (changes requested):
Can this be done using `EntityTable` and `useEntityTable`?

CodeNamedRobin (approved):

PR #858, opened 2026-08-04
Follow terminal payments over websocket instead of polling

JustSamuel:
# Description

> [!WARNING]
> **Do not merge until GEWIS/sudosos-backend#1011 is merged _and_ deployed to the target environment.** This removes polling entirely, so against a backend that does not emit `terminal_payment:updated` a payment will never settle and the cashier will wait forever. There is no fallback by design.

Stacked on #850 — review that first. The base is `feat/pos-terminal-payment`, so the diff here is only the two files this change touches.

The POS polled `GET /terminal-payments/{id}` every two seconds until a payment settled. Each poll logs a controller trace, so a thirty second payment produced roughly fifteen of them, and the cashier waited out an interval after the customer had already tapped.

It now subscribes to `terminal_payment:{id}:updates` and reacts to `terminal_payment:updated`.

Two single fetches remain, neither a loop:

- immediately after subscribing, because joining a room is asynchronous server-side and a contactless tap can settle a payment before the join completes
- after a reconnect, since anything emitted while the socket was down is gone for good

That takes a payment from about fifteen requests to one.

## Details worth a reviewer's eye

**Subscribing happens before the reader is engaged.** The backend emits `processing` from the same request that starts the payment, and a tap can settle it within a second, so listening afterwards can miss both.

**The room name uses underscores deliberately.** The backend parses rooms as `[a-z_]+:{id}:[a-z_]+`, so `terminal-payment:...` would fail to parse and nothing would ever be delivered. There is a test on the backend side asserting the hyphenated variant does not parse.

**`listenFor` reports failure by return value rather than by setting the phase.** The caller has to cancel the payment first, and cancelling sets the phase to `cancelled`, which would overwrite an error set inside the method.

**Authorisation works out because of which token the socket uses.** `apps/point-of-sale/src/utils/beforeLoadUtil.ts` connects with the POS token, which carries the `Point of Sale` role and its `TerminalPayment: get own` permission. An anonymous payment's `from` and `createdBy` are both the POS user, so the backend room policy resolves `own` and the subscription is authorised. Had the socket used the cashier's token it would have been rejected, since a cashier has no `TerminalPayment` permission at all.

## Testing

Verified: `vue-tsc --noEmit` on both apps, `eslint`, and `prettier --check`, all clean.

Not verified end to end, because the backend does not emit these events yet on any deployed environment. Once #1011 is deployed the flow to exercise is: tap a card and confirm the modal closes on the push rather than after a poll interval, cancel mid-payment and confirm the subscription is torn down, and drop the network briefly while waiting to confirm the reconnect fetch reconciles.

The store now has no timers, so the previous polling edge cases (consecutive-failure tolerance, the 403/404-means-cancelled special case) are gone from the waiting path. The 403/404 handling is retained on the reconcile fetch, since a Stripe-side cancellation still detaches the transaction and can drop the POS token out of the `own` relation.

Depends on GEWIS/sudosos-backend#1011, which closes GEWIS/sudosos-backend#1010. Stacked on #850.

## Types of changes

### DIFF
--- a/apps/point-of-sale/src/components/Cart/TerminalPaymentModal.vue
+++ b/apps/point-of-sale/src/components/Cart/TerminalPaymentModal.vue
@@ -171,7 +171,7 @@ const onCancel = async () => {
 };
 
 onUnmounted(() => {
-  terminalPaymentStore.stopPolling();
+  terminalPaymentStore.stopListening();
 });
 </script>
 
--- a/apps/point-of-sale/src/stores/terminalPayment.store.ts
+++ b/apps/point-of-sale/src/stores/terminalPayment.store.ts
@@ -1,5 +1,7 @@
 import { defineStore } from 'pinia';
 import type { StripePaymentTerminalResponse, TerminalPaymentResponse, TransactionRequest } from '@gewis/sudosos-client';
+import { useWebSocketStore } from '@sudosos/sudosos-frontend-common';
+import type { Socket } from 'socket.io-client';
 import { posApiService } from '@/services/ApiService';
 
 /**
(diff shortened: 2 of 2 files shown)


RubenLWF (approved):
ja mooi ofzo

PR #863, opened 2026-08-07
fix: modals closable with escape

RubenLWF:
# Description

Fixes pin terminal softlock that occurred when pressing the escape button when a payment was pending on the pin terminal. This would close the modal, but not cancel the transaction.

## Types of changes


JustSamuel (approved):
not sure what escape means but I believe

